PR #41, opened 2015-03-24
sqlite v3.8.8.3.

MariadeAnton:
Updated to sqlite v3.8.8.3

Previous was v3.8.8.2 but I just saw v3.8.8.3 was out so I went with the latest one


SRombauts:
Thanks a lot @MariadeAnton.

Issue #23, opened 2014-08-06
Column by name


gennesseaux:
Hi, tanks for your work.

Is there a way to get column by name instead of index.
Or could you add this to your todo list.

gennesseaux:
Thanks for your work.

PR #45, opened 2015-04-17
Fix issue: Column by name #23

HighSparrow:
add method `Column getColumn(const char* aName);` in Statement.h


SRombauts:
Thanks for this!

I will take care of the formatting issues revealed by cpplint in Continuous Integration.

Issue #48, opened 2015-05-05
Memory error in destructor


onuraslan:
I started to play with SQLiteCpp. I have a simple program:

    #include "SQLiteCpp/SQLiteCpp.h"
    
    int main (int argc, char *argv[])
    {
      SQLite::Database db("mydb.db", SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE);
      SQLite::Statement query(db, "INSERT INTO categories (title) VALUES (?)");
      query.bind(1, "example");
      query.exec();
    
      return 0;
    }
    
mydb.db and categories table exists with id and title cols. When I run this program with valgrind I am getting this error message:

    ==3857== Memcheck, a memory error detector
    ==3857== Copyright (C) 2002-2013, and GNU GPL'd, by Julian Seward et al.
    ==3857== Using Valgrind-3.10.0 and LibVEX; rerun with -h for copyright info
    ==3857== Command: ./a.out
    ==3857== 
    ==3857== Invalid read of size 4
    ==3857==    at 0x4E75CB0: sqlite3SafetyCheckSickOrOk (in /usr/lib/x86_64-linux-gnu/libsqlite3.so.0.8.6)
    ==3857==    by 0x4E80A72: sqlite3_errmsg (in /usr/lib/x86_64-linux-gnu/libsqlite3.so.0.8.6)
    ==3857==    by 0x4021DF: SQLite::Database::~Database() (Database.cpp:77)
    ==3857==    by 0x401CD0: main (in /tmp/a.out)
    ==3857==  Address 0x60e91d4 is 100 bytes inside a block of size 856 free'd
    ==3857==    at 0x4C29E90: free (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
    ==3857==    by 0x4E4DFED: sqlite3_free (in /usr/lib/x86_64-linux-gnu/libsqlite3.so.0.8.6)
    ==3857==    by 0x4E87E14: sqlite3Close (in /usr/lib/x86_64-linux-gnu/libsqlite3.so.0.8.6)
    ==3857==    by 0x4021C6: SQLite::Database::~Database() (Database.cpp:75)
    ==3857==    by 0x401CD0: main (in /tmp/a.out)
    ==3857== 
    ==3857== 
    ==3857== HEAP SUMMARY:
    ==3857==     in use at exit: 0 bytes in 0 blocks
[...]

PR #49, opened 2015-05-05
Use sqlite_errstr instead of sqlite3_errmsg

onuraslan:
mpSQLite is free'd with sqlite3_close. It's no more accesible.
Fixes: #48

### DIFF
--- a/src/Database.cpp
+++ b/src/Database.cpp
@@ -74,7 +74,7 @@ Database::~Database() noexcept // nothrow
 {
     const int ret = sqlite3_close(mpSQLite);
     // Never throw an exception in a destructor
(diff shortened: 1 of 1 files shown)


SRombauts:
Thanks a lot!

SRombauts:
Just to let you know, I had to revert this (I know use a constant string as error message) because sqlite3_errstr() is to new for good support under Linux : it is not in Ubuntu 12.04 used by Travis CI continuous integration.

onuraslan:
I see. Yes, it's much better to use a static message in this case.

SRombauts:
Well, at least it does the job.

PR #51, opened 2015-05-13
Added support for extension loading

portwaypoint:
I have added this to the loadExtension to Database class as preferred.

It passes all your existing unit tests, but doesn't have an explicit unit test, since this would require a dummy module to be added to the source base, which I didn't think was a good idea!

I have however tested against my application and it successfully loads and integrates a module into SQLite3.


SRombauts:
Thank you very much for your contribution!

PR #52, opened 2015-05-16
Added conditional support for extension loading

nazim-can-bedir:
SQLite extension loading feature can be disabled by defining [SQLITE_OMIT_LOAD_EXTENSION preprocessor macro](http://sqlite.org/compile.html#omitfeatures). If anyone defines SQLITE_OMIT_LOAD_EXTENSION, then compilation of Database.cpp fails with undefined external symbol. Database::loadExtension function needs conditional compile-time check for successful compilation in such environment.


SRombauts:
Thank you.

PR #57, opened 2015-05-21
Build as dll

aurelien35:
I've made following changes :
- added SQLiteCppExport.h header which contains DLL import / export directive for Windows and Linux
- added the SQLITECPP_DLL macro on all symbols to be exported
- modified the CMakeList file to add new options

There is 3 new preprocessor "DEFINE" symbols :
- SQLITECPP_DYNAMIC : 
  => if defined, this mean SQLiteCpp is being build/used as a dynamic library
- SQLITECPP_EXPORT: 
  => if defined, this mean SQLiteCpp is being BUILD as a dynamic library, and it will EXPORT its symbols
  => if not defined, this mean SQLiteCpp is being USED as a dynamic library, and it will IMPORT its symbols
- SQLITECPP_DISABLE_MSVC_DLL_WARNINGS :
  => if defined, this mean Visual Studio should ignore the DLL export warnings about STL classes. You can ignore that when you build all your solution with the same runtime library. (if you don't do that, you will have very bad time anyway)
# 

I'm not an expert in CMake files, and there is still something to fix in the actual file : the "SQLITECPP_EXPORT" is defined when a dynamic library build is asked, but even to build the example.

The example should not define this symbol because it need to IMPORT the code, not export it.

My guess is a subdirectory with a specific configuration is needed for every target : SQLite3, SQLiteCpp and Example

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,12 +13,16 @@ project(SQLiteCpp)
 if (WIN32)
     set(DEV_NULL                "NUL")
     # build the SQLite3 C library for Windows (for ease of use)
-    set(SQLITECPP_INTERNAL_SQLITE_DEFAULT     ON)
-    set(SQLITE_ENABLE_COLUMN_METADATA_DEFAULT ON)
+    set(SQLITECPP_INTERNAL_SQLITE_DEFAULT           ON)
+    set(SQLITECPP_BUILD_DYNAMIC_LIBRARY_DEFAULT     OFF)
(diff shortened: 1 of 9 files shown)


SRombauts:
Yes, you are right on your assumption that every target would need a specific CMakeLists.txt in its subdirectory.

I'll try to take the time to review this.

SRombauts:
Closing this as it would need more work that I afford

SRombauts:
This was implemented long after by #406

PR #67, opened 2015-09-23
Fix implicit conversion to std::string of Column class for Visual C++ 2015

HNO3:
I have changed `#if` condition masking `Column::operator std::string`, because a problem that implicit conversion to `std::string` does not work if there are both `operator std::string` and `operator const char *` is fixed from Visual C++ 2015.

The condition became more generic to support "standard" compilers not defining `__GNUC__`.


SRombauts:
Thanks for tour contribution!

PR #68, opened 2015-10-01
Removed unusel googletest module.

madmaxoft:
Removed a submodule that seems to have moved and is not needed for our compilation anyway.


madmaxoft:
Sorry, this wasn't meant to go in the upstream repo.

PR #85, opened 2016-06-15
Variadic bind convenience function

pauldreik:
This gives the possibility to let the compiler generate the index argument in `Statement::bind()` automatically at compile time, instead of manually invoking bind.
Instead of
`SQLite::Statement stmt("SELECT * FROM MyTable WHERE colA>? && colB=? && colC<?");
stmt.bind(1,a);
 stmt.bind(2,b);
 stmt.bind(3,c);
`
one can write
`SQLite::bind(stmt,a,b,c);`

Note that this is implemented as a free function. It could be provided as a member function of `SQLite::Statement` instead, under a different name than bind in order not to clash with the existing `SQLite::Statement::bind()` overloads.

This requires c++14. Tested on gcc 5.3.1 on GNU/Linux.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,6 +99,7 @@ set(SQLITECPP_TESTS
  tests/Database_test.cpp
  tests/Statement_test.cpp
  tests/Backup_test.cpp
+ tests/VariadicBind_test.cpp
 )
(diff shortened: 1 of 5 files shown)


SRombauts:
Thanks a lot for this high quality work!

PR #87, opened 2016-06-27
build with several compilers

pauldreik:
- multiple versions of gcc (default which is 4.8, 4.9, 5, 6)
 - multiple versions of c++ standard ( 03, 11, 14, 1z)
 - multiple versions of clang(default which is 3.5, 3.8)
 - os x

see this test build: [https://travis-ci.org/pauldreik/SQLiteCpp/builds/140468462](url)


SRombauts:
Hi Paul, thanks a lot for all your work, it is really appreciated!

Cheers!

PR #90, opened 2016-07-11
Automatically propagate public headers to other projects with CMake

4brunu:
I'm adding SQLiteCpp to my project with CMake, using `add_subdirectory`.

The problem I found is that I need to manually add the path to the public header files (SQLiteCpp/include/ and SQLiteCpp/sqlite3), since they are not propagated automatically.

This PR add the `target_include_directories` so that the projects that include SQLiteCpp don't need to include the path to the public header files manually.


SRombauts:
Hi @4brunu,

Thank you for this contribution, it is appreciated!

Cheers!

PR #91, opened 2016-07-15
Add option to check if column is null by name

4brunu:
While using the library I wanted to check is a column is null, but only found a method that allow to check it by index, so I added one that is similar but check by the column name.

I have also added the tests to it.

### DIFF
--- a/include/SQLiteCpp/Statement.h
+++ b/include/SQLiteCpp/Statement.h
@@ -435,7 +435,18 @@ public:
      *
      *  Throw an exception if the specified index is out of the [0, getColumnCount()) range.
      */
-    bool    isColumnNull(const int aIndex) const;
+    bool    isColumnNull(const int aIndex);
+
+    /**
+     * @brief Test if the column value is NULL
+     *
+     * @param[in] apName   Aliased name of the column, that is, the named specified in the query (not the original name)
+     *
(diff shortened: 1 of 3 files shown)


SRombauts:
Thanks @4brunu!

I merged it as-is, but I will rework some of it:
- removing const on isColumnNull() does not seems right
- I believe duplicating code to map indexes to names is not a good idea either

Cheers!

4brunu:
Hi @SRombauts, sure, no problem at all :)
 
Just to elaborate on some of the decisions:
- I removed const because it was giving a compiler error. I didn't understand why, could you elaborate on that please?
- I tried to do that, but I ended up reverting that because it was resulting in some issues, due to my lack of c++ experience.

Thanks for the amazing library 👍

SRombauts:
The const compiler error was weird: the problem is that we know have two overloaded functions isColumnNull(), one taking an (int), the second one taking a pointer (const char*).
Thus when calling it with zero isColumnNull(0) the compiler has to decide witch one is the best one since 0 is also a nullptr.

There is no problem if the two methods are both const or not const, but there is an error if only one is const: I think that the compiler is not able to decide witch one is the best are they are not real overload...

4brunu:
Thanks a lot for the explanation :)

PR #96, opened 2016-08-04
Fixed link errors calling Exception::getErrorCode()

snej:
getErrorCode() and getExtendedErrorCode()'s implementations were
accidentally declared as inline in the .cpp file. This causes the
compiler to not generate any code for them, resulting in link errors
when a client calls them. (Observed compiling with Clang.)

Fixed by moving the implementations into the header, where they need to
be if they're inline.


SRombauts:
Thanks for this catch, I'll update unit-tests to cover these cases

PR #102, opened 2016-09-09
Redundant const specifier removed

Nemo1369:
Fixes clang warning on Database.h:48: "'const' type qualifier on return type has no effect" and Database.cpp:44: "'const' type qualifier on return type has no effect"


SRombauts:
Thanks @Nemo1369, good catch!

PR #107, opened 2016-12-01
Added ability to open encrypted databases.

jowr:
No advanced functionality, but one can open encrypted databases. requires you to have an external SQLite library that supports encryption.

### DIFF
--- a/include/SQLiteCpp/Database.h
+++ b/include/SQLiteCpp/Database.h
@@ -373,10 +373,8 @@ public:
     * @brief Set the key for the current sqlite database instance.
     *
     *  This is the equivalent of the sqlite3_key call and should thus be called 
-    *  directly after opening the database. If the database is unencrypted, 
-    *  this methods encrypts it immediately.
-    *  Normal database -> call db.key("secret") -> encrypted database, database ready
-    *  Encrypted database -> call db.key("secret") -> database ready
+    *  directly after opening the database. 
+    *  Open encrypted database -> call db.key("secret") -> database ready
     *
     * @param[in] aKey   Key to decode/encode the database
     *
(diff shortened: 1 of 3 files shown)


SRombauts:
Hi, thanks for provinding this!

I am not entirely sure on how to integrate your code. I would prefer not to add more parameters to the Database constructor since there are already to much of them.

Would it be possible to use a setter instead, a Databse::setKey() for instance?
This would better stick with the underlying SQLite3 API.

Cheers!

jowr:
Hi, it might be possible. The only problem is that **all** queries on encrypted databases fail until one calls `sqlite3_key` with the correct password. This might cause `setBusyTimeout` to fail, but I have not tested that. That is why I added before that call.

SRombauts:
Hi, nice to see you working on this!
À few issues left with gcc/clang under Linux though.

jowr:
I think we are there now, feel free to merge it, but I did not test it with `SQLITE_HAS_CODEC` enabled on other platforms than Windows...

SRombauts:
Thanks a lot for your work!

include/SQLiteCpp/Database.h
```diff
@@ -369,6 +369,51 @@ class Database
...
+    *  header does not match the standard string, we assume that we have an 
+    *  encrypted file. 
+    *
+    * @param[in] aFilename path/uri to a file
+    *
+    * @return true if the database has the standard header.
+    *
+    * @throw SQLite::Exception in case of error
+    */
+    static const bool isUnencrypted(const std::string& aFilename);
```
timrae: `const bool` return type results in the following gcc warning:
`type qualifiers ignored on function return type`

Since you are returning by value the `const` is unnecessary.

timrae: Patch submitted in #119

jowr: Thanks for reviewing the code!

PR #108, opened 2016-12-05
Use ExternalProject for googletest

timrae:
There were some TODO comments about using the official github directly from the cmake file, and due to some issues with cmake I can't use projects with submodules in my organization, so I went ahead and updated the cmake file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,7 @@
 # or copy at http://opensource.org/licenses/MIT) 
 cmake_minimum_required(VERSION 2.8.12) # first version with add_compile_options()
 project(SQLiteCpp)
+include(ExternalProject)
 
 # Define useful variables to handle OS differences:
 if (WIN32)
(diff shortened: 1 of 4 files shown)


SRombauts:
Thanks a lot for provinding this! 
Please have a look at the build errors, something is not working as intended...

timrae:
Yeah it's the yml files, I'll fix them now

timrae:
Ugh, it looks like it's still failing on the Windows builds because I used `pthread`. I'll look into what the multi-platform lib should be

timrae:
I can't see the error on travis, can you paste it here?

CMakeLists.txt
```diff
@@ -249,22 +250,31 @@ endif (SQLITECPP_BUILD_EXAMPLES)
...
+        GIT_REPOSITORY https://github.com/google/googletest.git
+        INSTALL_COMMAND ""
+        CMAKE_ARGS -DCMAKE_C_COMPILER=$ENV{CC} -DCMAKE_CXX_COMPILER=$ENV{CXX}
+    )
 
     # add the unit test executable
     add_executable(SQLiteCpp_tests ${SQLITECPP_TESTS})
-    target_link_libraries(SQLiteCpp_tests gtest_main SQLiteCpp sqlite3)
+    add_dependencies(SQLiteCpp_tests googletest)
+    set(GTEST_MAIN_OBJ ${PREFIX_DIR}/src/googletest-build/googlemock/gtest/${LIB_PREFIX}gtest_main${LIB_SUFFIX})
```
timrae: @SRombauts 
Sorry I don't have access to a Windows C++ environment, so I don't know where these gtest binaries are being generated and what they're called there. On linux they're called `libgtest.a` and `libgtest_main.a`. Would you be able to check what's happening on Windows?

SRombauts: Sure, I'll try this at some point.

SRombauts:
Closing, would need some serious rework

PR #111, opened 2016-12-28
linked SQLiteCpp to sqlite3

Yanpas:
Now there is no need to link each target ot both libs


4brunu:
This would be great.
@SRombauts What do you think of this?

SRombauts:
Hi @Yanpas,

Sorry for the very long reaction time, I had to look to documentation, and I added a comment in the CMakeLists.txt file.

It should be easier to use the wrapper now, thanks a lot!

Cheers

SRombauts:
Hi @Yanpas, sorry but I'll have to revert this since it is interacting badly with the new "install step" from #118

timrae:
Probably you need to do this inside the `if (SQLITECPP_INTERNAL_SQLITE)` block... I haven't tested it yet though.

Also pinging @4brunu

4brunu:
@timrae I tried that, but it return's the same error
https://travis-ci.org/4brunu/SQLiteCpp/jobs/221700112
Any more suggestions?

PR #114, opened 2017-02-13
Added convenience functions for constructing objects from a row

dend456:
Added a template function that allows you to retrieve all columns (or first N columns) of a row as an object. Extremely convenient when working with a table that has a large number of columns. Instead of having to do:
int i0 = query.getColumn(0);
int i1 = query.getColumn(1);
...
int i39 = query.getColumn(39);

then passing all those temporary variables into a constructor, you can do:

auto i = query.getColumns<MyClass, 40>();

where MyClass has a constructor that takes 40 arguments that are convertible from a Column object.

### DIFF
--- a/include/SQLiteCpp/Column.h
+++ b/include/SQLiteCpp/Column.h
@@ -260,5 +260,22 @@ private:
  */
 std::ostream& operator<<(std::ostream& aStream, const Column& aColumn);
 
+#if __cplusplus >= 201402L || (defined(_MSC_VER) && _MSC_VER >= 1900)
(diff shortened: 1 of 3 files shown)


SRombauts:
Hi,

Thanks a lot, this is great!

dend456:
Np, do you think a function that fills a std::vector<T> from all rows in the result would be useful?

SRombauts:
Well, it could, but the problem are:
1. you would like to be able to use any container
2. I am often using container of unique or shared ptr

PR #116, opened 2017-03-15
added private assigment operator to Exception class to avoid MSVC warning

Philipp2405:
added private assigment operator to Exception class to avoid MSVC warning: C4512: 'SQLite::Exception' : assignment operator could not be generated ...' --> only created at warning level 4 !

### DIFF
--- a/include/SQLiteCpp/Exception.h
+++ b/include/SQLiteCpp/Exception.h
@@ -91,6 +91,9 @@ public:
     const char* getErrorStr() const noexcept; // nothrow
 
 private:
+    // due to constant members the auto-assignment operator can not be generated
+    Exception& operator=(const Exception&);
+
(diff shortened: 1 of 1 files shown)


SRombauts:
Hi, sorry for not answering, I am not sure that this is the best fix for this warning...

fekir:
HI,

subclasses of std::exception should have an assignment operator (I would actually expect that every class that has a copy constructor also to have an assignment operator).

I've added such operator in PR https://github.com/SRombauts/SQLiteCpp/pull/138, this should also fix the warning of Philipp2405

SRombauts:
Thanks @fekir!

As PR #138 is merge I can now close this. Sorry for the delay @Philipp2405

Cheers!

PR #118, opened 2017-03-27
Add install step

timrae:
Currently, use of the library is quite cumbersome for us because it's lacking an install step.

This PR ensures that `make install` works correctly, and that users can reliably find it using `find_package` after `make install` has completed without any additional files or work.


SRombauts:
Hi, thanks for this!

SRombauts:
Hey @timrae, could you please have a look of the issue with this "install" step I have just discovered?

Cheers!

timrae:
Ok I'll try and take a look!

timrae:
Does the issue only occur with #111 or even without it?

SRombauts:
The error occurs only with #111 si I commented if out temporarily

timrae:
I think it's more usual for library users to link in both libraries, no?

PR #119, opened 2017-03-31
Fix some warnings and increase safety checking on gcc

timrae:


timrae:
I'd recommend adding the `-Werror` flag as well so that the CI won't pass if there are any warnings

SRombauts:
Hi, thanks for this. 

I am pretty sure I started this project with -Wextra, I don't know when I would have lost it...

PR #120, opened 2017-04-06
Make cpplint.py Python-3 compatible

### LINKED ISSUE #99: Project doesn't build on arch linux.
Project doesn't build on arch linux.

Arch linux uses python3 as default python interpreter. But cpplint.py seems to require python2. Although running it without any parameter it will print out help message happily but when used with parameters it will crash without any message.

you should add something in the CMakelists.txt, perhaps before find_package(PythonInterp).

vbraun:
Smallest fix to let cpplint.py work on python2 and 3

FIxes #99

### DIFF
--- a/cpplint.py
+++ b/cpplint.py
@@ -53,6 +53,15 @@ import sys
 import unicodedata
 
 
+try:
+    xrange(0,1)
+    PY3 = False
+except NameError:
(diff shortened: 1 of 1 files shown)


SRombauts:
Thanks !

PR #136, opened 2017-08-18
Ensure that operator<< correctly prints strings with embedded '\0'

fekir:
Hello,

std::ostream& operator<<(std::ostream& aStream, const Column& aColumn) does not handle correctly content with an embedded '\0'.
I've also added a test case with a std::string.

Best regards
Federico


SRombauts:
Thanks again for you work Federico!

Regards,
Sébastien

PR #137, opened 2017-08-19
Remove noexcept

fekir:
Hi,

there are a couple of method marked as noexcept, but in reality they may throw an exception.

The first method is Column::getString, since std::string needs to allocate memory, the constructor may fail and throw an exception

The second method is Database::setBusyTimeout, internally if the operation fails an exception is thrown.

I was going to remove the noexcept specifier from the destructors, since they are all noexcept if not stated otherwise, but maybe you preferred to state it explicitely, and since it does not make any difference from the language perspective, I leaved them.

### DIFF
--- a/include/SQLiteCpp/Column.h
+++ b/include/SQLiteCpp/Column.h
@@ -111,7 +111,7 @@ public:
      *
      * Note this correctly handles strings that contain null bytes.
      */
-    std::string getString() const noexcept; // nothrow
+    std::string getString() const;
 
     /**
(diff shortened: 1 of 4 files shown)


SRombauts:
Thanks @fekir, I am a bit ashamed by the setBusyTimeout() mistake!

fekir:
Would you accept a pull request that removes the unnecessary noexcept from destructors (and nothrow comment)?

SRombauts:
gladly!

PR #138, opened 2017-08-19
Enhance sqlite exception

fekir:
Hi,

subclasses of std::exception should have an assignment operator.

I've added test cases to ensure that such operator behaves correctly.

I've also added two constructor that takes const char* instead of const std::string& in order to avoid a possible allocation (in c++11) when creating the exception object.


SRombauts:
Hi @fekir, thanks for this!

PR #139, opened 2017-08-21
Remove unnecessary noexcept identifier from destructors

fekir:
This is a pull request for "cleaning" the destructors.

I've removed the noexcept identifier since it is unnecessary.

Some destructor are empty, I leaved them since most where commented.

I've noticed that they are all declared virtual, but there are no virtual method in any class or any subclass. 
Is there a reason? Otherwise we could also remove the virtual identifier.

### DIFF
--- a/include/SQLiteCpp/Backup.h
+++ b/include/SQLiteCpp/Backup.h
@@ -100,7 +100,7 @@ public:
            Database& aSrcDatabase);
 
     /// Release the SQLite Backup resource.
-    virtual ~Backup() noexcept;
+    virtual ~Backup();
 
     /**
      * @brief Execute a step of backup with a given number of source pages to be copied
--- a/include/SQLiteCpp/Column.h
+++ b/include/SQLiteCpp/Column.h
@@ -54,7 +54,7 @@ public:
      */
     Column(Statement::Ptr& aStmtPtr, int aIndex)    noexcept; // nothrow
     /// Simple destructor
-    virtual ~Column()                               noexcept; // nothrow
+    virtual ~Column();
 
     // default copy constructor and assignment operator are perfectly suited :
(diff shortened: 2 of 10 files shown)


SRombauts:
Thanks!

SRombauts:
Virtual dtor are there so that you can specialize any class, but I am now unsure if it have any meaning.
Perhap's did I leave a clue in the git log when I added those "virtual" and the comments about "no virtual dtor needed here" ?

I would need to dive into this specific subject to be certain, but I lack the time right now. If you want to look at this, and have a discussion about it, feel free!

fekir:
I gave a quick look at git log, the Database class destructor has been virtual from it's first commit (commit 733c92e753752d757111306ee7c9024ff843955d), and the Ptr class always had that comment (commit 1a825108219f9d324afe969d2c835c0595094091).

I do not think that you can specialize anything since there are no virtual methods.
I am not  a big fan of class hierarchies, thats why I immediately noticed the "virtual" keyword.

Since no one used it till today, I think there is no need for it.

Should we maybe move the discussion to a separate issue?

SRombauts:
Thanks for looking at it. I fell like you (now) about class hierarchy, I may have been otherwise at that time..

So yes you can remove it, I will make sure to mention it on the changelog since it may break someone else code in a subtle way.

fekir:
I've pushed my changes to my branch "clean_destructors" (I accidentally deleted it, but was able to restore it). 

Are you able to merge it again or do I need to make a new pull request?
It seems I need to do a new since the other time I've updated the branch (before you merged it back), the ci system build everything again, and it does not seem to happen now.

I hope it is possible to make a request for the same branch again...

SRombauts:
You can probably make a new merge request, but this one cannot be re-used for another purpose.

PR #140, opened 2017-08-22
Remove virtual destructor since there are no virtual methods or subclasses

fekir:
As stated in https://github.com/SRombauts/SQLiteCpp/pull/139, virtual destructors have (currently) no use case


fekir:
I've already removed the comment when I removed the nothrow statement and throw() comment

SRombauts:
Thanks again!

PR #148, opened 2017-11-09
Allows long int for bind when used with name

xforce:
a3160dcfc2f80f692f3477a67c202ff87f75fa5a fixed this for bind by index

This PR just applies the same fix to the named functions.


SRombauts:
Well, sure, I was too quick and missed that!
Thanks for doing this, with proper unit tests.

PR #151, opened 2017-12-15
more cmake instructions for linux

gocarlos:
adding only SQLiteCpp to target_link_libraries() was not enough for me under ubuntu 16.04. 
there fore added cmake instructions in more detail.


SRombauts:
Thanks @gocarlos, this kind of enhancement to documentation is always helpful!

PR #157, opened 2018-03-29
Add a move constructor to Database

hubslave:
This makes it possible to e.g. return Databases from functions. Gated behind a __cplusplus >= 201103L check for compatibility with older C++ versions. Defined inline due to its triviality and in order to avoid having to do the same preprocessor dance in two places.

Despite setting mpSQLite to null, there is no need to change the destructor, as sqlite3_close handles null pointers for us.


SRombauts:

Cheers!

PR #170, opened 2018-12-03
CMake compatibility

jowr:
- Added backwards compatibility with older CMake-versions. 
- Fixed a problem with the constantly failing encryption tests

There are quite a few non-sense commit messages, please squash-merge to master and write a single meaningful commit message.


SRombauts:
Thanks for your work on this!

PR #171, opened 2018-12-12
Variant compatibility

JakubFornadel:


SRombauts:
Seems interesting to me. Why did you close? Is it not working as expected?

JakubFornadel:
it is working but it is using variant from fc library. This library is pretty big with a lot of stuff inside so it is not very suitable for your SQLiteCpp i guess...

SRombauts:
Understood, thanks

PR #172, opened 2018-12-18
Update VariadicBind.h

JakubFornadel:
Remove warning: expression result unused [-Wunused-value]
https://stackoverflow.com/questions/14370043/can-operators-be-overloaded-for-initializer-list-literals


SRombauts:
Thanks

PR #176, opened 2019-02-01
Fix error message regarding git submodule init/update.

tjhowse:
This error message suggests incorrect git commands to init and update the required submodule/s. This PR corrects the suggestions.


SRombauts:
Thank you!

PR #192, opened 2019-03-04
Add wrapper for bind parameter count

jrave:


SRombauts:
Nice addition to the wrapper, thank you!

PR #194, opened 2019-03-30
Fix problem with MSVC 19

### LINKED ISSUE #193: error C2039: '_snprintf': is not a member of 'std' - Windows MSVC 19
Hi
A combination of SQLiteCpp with boost 1.69.0 and MSVC 19.16.27027.1 gives me following error:
```
c:\.conan\a15976\1\include\boost\system\detail\system_category_win32.hpp(52): error C2039: '_snprintf': is not a member of 'std' 
```
The code causing this issue seems to be these three lines:
https://github.com/SRombauts/SQLiteCpp/blob/50425142fec09849c8114f8a042d72f5bcf7f5b5/include/SQLiteCpp/Utils.h#L67-L69

Changing the code like suggested [here](https://github.com/nlohmann/json/issues/1408#issuecomment-450559270) should fix the problem:
```
#if defined(_MSC_VER) && _MSC_VER < 1500 // VC++ 8.0 and below
#define snprintf _snprintf
```
I will create a PR soon...

Linux13524:
fixes #193

### DIFF
--- a/include/SQLiteCpp/Utils.h
+++ b/include/SQLiteCpp/Utils.h
@@ -64,6 +64,6 @@ const nullptr_t nullptr = {};
 #endif // _MSC_VER
 
 // A macro for snprintf support in Visual Studio
-#if _MSC_VER
+#if defined(_MSC_VER) && _MSC_VER < 1500
 #define snprintf _snprintf
 #endif


SRombauts:
Thanks for the fix, nice job!

PR #196, opened 2019-05-09
Update VariadicBind.h

maxbachmann:
I had to write some sqlite stuff recently and found this library afterwards, so I checked how the variadic bind is implemented here.
This implementation is simpler and does not require any C++14 functionality

### DIFF
--- a/include/SQLiteCpp/VariadicBind.h
+++ b/include/SQLiteCpp/VariadicBind.h
@@ -44,11 +44,10 @@ template<class ...Args>
 void bind(SQLite::Statement& s, const Args& ... args)
 {
     static_assert(sizeof...(args) > 0, "please invoke bind with one or more args");
-    
-    int pos = 1;
+    int pos = 0;
 	(void)std::initializer_list<int>{
-		((void)s.bind(pos++, std::forward<decltype(args)>(args)), 0)...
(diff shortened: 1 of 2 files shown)


maxbachmann:
Fixed ;)

SRombauts:
Thanks a lot for this contribution!
I had no time to review it, but will try to do this in the coming weeks (please get back to haunt me if I don't!

SRombauts:
Hi, could you please also update unit tests to now run on c++ 11

maxbachmann:
totally forgot them ;)

maxbachmann:
Oh I guess I have some stuff to fix

maxbachmann:
@SRombauts in the tests there is the following comment

> zero arguments - should give compile time error through a static assert
   SQLite::bind(query);

however this is not tested. I removed this assertion in my code aswell, since when no arguments are given the function will just do nothing. (the bind functions will just not get called). 
is there any reason why this assertion would be required?

SRombauts:
You are right, it read the same comment yesterday, and also wanted to get rid of it.

SRombauts:
Thank you again!

I squashed everything into one commit to try keeping the history a bit clearer to the readers.

Issue #24, opened 2014-08-19
Dealing with multiple inserts


saaqibz:
Hello, Is there a command like executemany() for dealing with multiple inserts?

SRombauts:
For future reference, [link to the python documentation](https://docs.python.org/2/library/sqlite3.html):

```python
# Larger example that inserts many records at a time
purchases = [('2006-03-28', 'BUY', 'IBM', 1000, 45.00),
             ('2006-04-05', 'BUY', 'MSFT', 1000, 72.00),
             ('2006-04-06', 'SELL', 'IBM', 500, 53.00),
            ]
c.executemany('INSERT INTO stocks VALUES (?,?,?,?,?)', purchases)
```

xsacha:
If you were to make this function, would it be useful to have a bool option to start transaction and end transaction around the execution?
Right now I have a function myself that does this once per second.

SRombauts:
Hi @xsacha, I am not sure of what you mean, can you give a simple example?

xsacha:
I just mean to have:

db->exec("BEGIN");
query->reset();
for (auto statement: collection)
{
    // bind everything in statement
    query->bind(..);
}
db->exec("END");

SRombauts:
This as finally been implemented and integrated, thanks to @maxbachmann!

PR #197, opened 2019-05-16
add tuple bind

maxbachmann:
added the option to use a tuple with values to bind.
I will use this to implement executeMany as described in  #24, so it will be possible to pass multiple tuples

I only created this PR, so the CI already tests this, since I still need to implement executemany and add some tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ build
 example1
 *.a
 
+.vscode/
 /SQLiteCpp.sln
 *.ncb
 *.suo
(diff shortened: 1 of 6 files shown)


maxbachmann:
I did not test this at all yet, so I guess it should work, but I definetly need to write a unit test for these functions aswell and add some comments. However when there is already something you would do in a different way feel free to comment on it ;)

maxbachmann:
@SRombauts I added test cases, comments and fixed a few broken things, so now it should be ready for review and integration

maxbachmann:
@SRombauts little reminder :)

SRombauts:
Thank you very much for this contribution!

I was very busy recently (releasing a pre alpha version of a game) but I will try to also review and merge the few other interesting PR waiting.

Cheers!

Issue #156, opened 2018-03-26
Misleading error message in exception from Statement::exec


alanbirtles:
If Statement::exec is executed with mbDone set to true then an exception is thrown with the correct error code SQLITE_MISUSE returned from tryExec but the error message is "not an error"

SRombauts:
Good catch, thanks for reporting it!

tiendq:
@alanbirtles How does it happen? Did you mention [this function](https://github.com/SRombauts/SQLiteCpp/blob/master/src/Statement.cpp#L257)?

alanbirtles:
just execute a statement twice without resetting it

PR #199, opened 2019-05-21
Fix Exception

maxbachmann:
Fix issue #156
the problem is that tryExecuteStep returns SQLITE_MISUSE when it was not used properly. Since this is set manually this is not the error state of the statement, so when checking the error message of the statement there obviously is none, since there was no error. This problem can not only occur in the exec function, but in executeStep aswell when used improperly. This PR fixes this problem for both functions by checking whether the error code is the same as the error state of the statement


SRombauts:
Thanks again, looking good!

PR #201, opened 2019-06-15
Add Statement::getExtendedSQL()

os12:
- it returns a UTF-8 string containing the SQL text of prepared statement with
   bound parameters expanded

PR #205, opened 2019-06-22
Fix a memory leak that came with the Statement::getExtendedSQL() API

### LINKED ISSUE #203: Memory leaks reported by the ASan build
I'm running my tests and the gcc/ASan build reports the following memory leaks:

```
==76229==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 20016 byte(s) in 253 object(s) allocated from:
    #0 0x7f0f79d16d28 in malloc (/usr/lib/x86_64-linux-gnu/libasan.so.3+0xc1d28)
    #1 0x563ba0ef95b3 in sqlite3MemMalloc (/home/oleg/logger/_obj/services/oracle/tests/oracle_tests+0x7a05b3)

Direct leak of 2640 byte(s) in 6 object(s) allocated from:
    #0 0x7f0f79d17090 in realloc (/usr/lib/x86_64-linux-gnu/libasan.so.3+0xc2090)
    #1 0x563ba0ef9558 in sqlite3MemRealloc (/home/oleg/logger/_obj/services/oracle/tests/oracle_tests+0x7a0558)

SUMMARY: AddressSanitizer: 22656 byte(s) leaked in 259 allocation(s).
```

It appears that they are coming `sqlite3MemMalloc ` which, of course is called from my code via SQLiteCpp. Given that SQLite itself is tested to an insane extent, I wonder whether something SQLite-specific leaks from the wrapper classes?

Could you take a look please? That is, could you build your tests with ASan and see please?

For GCC, it comes down to the following flags:
```
        CFLAGS="$CFLAGS -fsanitize=address -fno-omit-frame-pointer"
        CXXFLAGS="$CXXFLAGS -fsanitize=address -fno-omit-frame-pointer"
        LDFLAGS="-fsanitize=address"
```

os12:
- this block must be explicitly free'd, even though other similar things
   are automatically handled during statement finalization

https://www.sqlite.org/c3ref/expanded_sql.html

Fix #203

### DIFF
--- a/src/Statement.cpp
+++ b/src/Statement.cpp
@@ -416,7 +416,10 @@ const char* Statement::getErrorMsg() const noexcept // nothrow
 
 // Return a UTF-8 string containing the SQL text of prepared statement with bound parameters expanded.
 std::string Statement::getExtendedSQL() {
-    return sqlite3_expanded_sql(mStmtPtr);
+    auto statement = sqlite3_expanded_sql(mStmtPtr);
+    std::string rv = statement;
+    sqlite3_free(statement);
+    return rv;
 }
 
 ////////////////////////////////////////////////////////////////////////////////


maxbachmann:
you should not use the auto keyword in this place, since it requires c++11

SRombauts:
Thank you to you two.

For now it makes no sense to drop support for older compiler just for this sort of use of auto.

SRombauts:
For reference, this bug was introduced only a few days ago in #201

SRombauts:
I implemented Address Sanitizer to reproduce the problem, and fixed it on master.

Thanks again

Issue #206, opened 2019-06-24
Statement executeMany not compiling with GCC 9 C++17


SRombauts:
See https://travis-ci.org/SRombauts/SQLiteCpp/jobs/549915202

```
In file included from /usr/include/c++/9/pstl/glue_algorithm_defs.h:13,
                 from /usr/include/c++/9/algorithm:71,
                 from /home/travis/build/SRombauts/SQLiteCpp/googletest/googletest/include/gtest/internal/gtest-port.h:274,
                 from /home/travis/build/SRombauts/SQLiteCpp/googletest/googletest/include/gtest/internal/gtest-internal.h:40,
                 from /home/travis/build/SRombauts/SQLiteCpp/googletest/googletest/include/gtest/gtest.h:58,
                 from /home/travis/build/SRombauts/SQLiteCpp/tests/ExecuteMany_test.cpp:17:
/usr/include/c++/9/functional: In instantiation of ‘std::_Bind<_Functor(_Bound_args ...)>::_Bind(const _Functor&, _Args&& ...) [with _Args = {std::tuple<int>}; _Functor = SQLite::Statement; _Bound_args = {std::tuple<int>}]’:
/usr/include/c++/9/functional:811:38:   required from ‘typename std::_Bind_helper<std::__is_socketlike<_Func>::value, _Func, _BoundArgs ...>::type std::bind(_Func&&, _BoundArgs&& ...) [with _Func = SQLite::Statement&; _BoundArgs = {std::tuple<int>}; typename std::_Bind_helper<std::__is_socketlike<_Func>::value, _Func, _BoundArgs ...>::type = std::_Bind<SQLite::Statement(std::tuple<int>)>]’
/home/travis/build/SRombauts/SQLiteCpp/include/SQLiteCpp/ExecuteMany.h:84:9:   required from ‘void SQLite::bind_exec(SQLite::Statement&, std::tuple<_Tps ...>&&) [with Types = {int}]’
/home/travis/build/SRombauts/SQLiteCpp/include/SQLiteCpp/ExecuteMany.h:50:14:   required from ‘void SQLite::execute_many(SQLite::Database&, const char*, Arg&&, Types&& ...) [with Arg = std::tuple<int>; Types = {std::tuple<int, const char*>, std::tuple<int, const char*>}]’
/home/travis/build/SRombauts/SQLiteCpp/tests/ExecuteMany_test.cpp:35:9:   required from here
[...]

PR #207, opened 2019-06-25
improve execute many

maxbachmann:
execute many can handle different types now aswell and not only tuples.
This way it does not require the user to create a tuple to bind a single value:
```cpp
execute_many(db, "INSERT INTO test VALUES (?, ?)",
            1,
            std::make_tuple(2),
            std::make_tuple(3, "three")
        );
```

currently this means it supports tuples and single values when only one value has to be bound, since these are the only types supported by VariadicBind right now

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -198,31 +198,30 @@ matrix:
         - CC=gcc-9
         - CXX=g++-9
         - CXXFLAGS="-std=c++14 -Wall -Wextra -pedantic"
-# TODO: issue with ExecuteMany and Bind helper function
-#    # gcc 9 std=c++17
-#    - compiler: gcc
-#      addons:
-#        apt:
-#          sources:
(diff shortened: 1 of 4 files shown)


maxbachmann:
fixed #206

maxbachmann:
@SRombauts it's fixed now. However we should think about changing the name of the bind function to something like sql_bind so it is no name of a standard function anymore in the next major release

SRombauts:
Yes please, if you can work on this, it would be nice to improve the naming before I release a new version.

maxbachmann:
You could open a 3.x branch

SRombauts:
Thanks, I have integrated your fix!

SRombauts:
Regarding a 3.x branch, I am not yet ready to dedicated enough time on this

PR #211, opened 2019-07-06
Implement Database::backup()

os12:
- it is possible to easily backup/restore a live DB via the
   sqlite3_backup_init() API
   - https://www.sqlite.org/backup.html
   - https://www.sqlite.org/c3ref/backup_finish.html#sqlite3backupinit

### DIFF
--- a/include/SQLiteCpp/Database.h
+++ b/include/SQLiteCpp/Database.h
@@ -443,7 +443,8 @@ public:
      *
      * @return SQLITE_OK on success or an error code from SQLite.
      */
-    int loadOrSaveDb(const char* zFilename, int isSave);
+    enum class BackupType { Save, Load };
+    int backup(const char* zFilename, BackupType type);
 
 private:
     /// @{ Database must be non-copyable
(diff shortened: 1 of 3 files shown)


SRombauts:
Thanks a lot, I really like that you are making all these improvements! This one is even something I had implemented locally for one project, but never put int this wrapper.

I would suggest two changes though:

1. I usually try to stick with the underlying SQLite API name, so in this case it should be simply called "backup()"
2. The second parameter should be an enum, so it would be so much clearer at the call site without comments

If you agree to these I'll let you make the changes, but else QI am willing to merge this like that and improve on later myself

os12:
I took the name and the API verbatim from the sample... but sure, tweaked the names as per your suggestion. That took about a minute :)

SRombauts:
Thank you for your work, once again very appreciated!

I merged it into master and pushed a compilation fix for C++98, needed until we can start a new C+11 (or C++14) branch.

Cheers,
Sébastien

os12:
Cool! Thanks for the fix Sébastien!

Now I need to decide whether I should rebase :)

PR #212, opened 2019-07-22
fix set sqlite3 propertie

jiangzhuti:


SRombauts:
Hi!

Thanks for this, but can you explain the logic? I think we might actually want both sqlite3 and SQLiteCpp as dynamic libraries, depending on the use cases?

jiangzhuti:
> Hi!
> 
> Thanks for this, but can you explain the logic? I think we might actually want both sqlite3 and SQLiteCpp as dynamic libraries, depending on the use cases?
https://github.com/SRombauts/SQLiteCpp/blob/master/CMakeLists.txt#L192  VS  https://github.com/SRombauts/SQLiteCpp/blob/master/sqlite3/CMakeLists.txt#L15

SRombauts:
Thanks

PR #215, opened 2019-08-13
Disable implicit fallthrough warning when building internal sqlite3

zxey:
On gcc 7.3 or so, if we have -Wextra enabled, this warning also gets triggered. We cannot do anything about it, so just at least suppress the warning.


SRombauts:
Hi, thanks for the suggestion, but have you seen that it does not compile on older gcc version?

zxey:
Yeah, you are right, changed it so the warning is disabled only on gcc7 and greater where it is available.

SRombauts:

Sébastien

PR #219, opened 2019-09-04
Disable cast-function-type warning when building internal sqlite

zxey:
Disables cast-function-type warning which is available in gcc 8.0 and greater.


SRombauts:

Issue #103, opened 2016-09-17
dynamically create 'DB' item and use it until close


gencer:
On native sqlite3 library i can assign db object to a class variable and use it in other methods without open everytime.

But with SQLiteC++, I cannot share the object. Assertion or memory errors occured.

All I want create SQLite::Database and assign it to class so the whole class can use the db as needed without explicitly open everytime.

SRombauts:
Hi, I am not sure of what is the problem you get. Can you share the exact log of assertion or the type of memory errors ? 

Are you sharing the Database accross multiple threads? This is not allowed by SQLite itself (at least not with default compile options)

gencer:
No, Not like that. Let me explain;

```
class A {
public:
  static void openDb();
  static void insertValue(const char*);
private:
  static SQLite::Database db;
};

void A::openDb() {
  //something like this->db
  db(db_path, params)
}

void A::insertValue(const char *value) {
  //this->db or just db or A::db...
  SQLite::Statement query(db, "insert into...");
}
```

SRombauts:
Okay, what I understand is that you want to open the db dynamically (that is, not in the constructor like in my examples) and then use it many times until closes.

Best way to do this would be to use a pointer to Database. 

Let me get back home to make an example for you.

gencer:
@SRombauts exactly! Thank you for the example. I am awaiting your reply on this.

_BTW, For this case, which is more stable or proper? Opening every time and close at the end of execution? Or use it until app gets closed?_

SRombauts:
So, first of all, here is how I use SQLiteCpp usually, with constructors (RAII):

```C++

class A {
public:
    A(const char* path);
    void createTable();
    void insertValue(const char* value);
private:
    SQLite::Database    db;
    SQLite::Statement   insert;
};

A::A(const char* path) :
    db(path),
    insert(db, "INSERT INTO test VALUES (NULL, ?)")
{
}

void A::insertValue(const char* value)
{
    insert.bind(value);
    insert.exec();
}

int main ()
{
    A a("test.db3") // "CREATE TABLE test (id INTEGER PRIMARY KEY, msg TEXT)";
    a.insertValue("example");
}
```

SRombauts:
And then, an example with dynamic allocation, using shared pointers:

```C++

#include <SQLiteCpp/SQLiteCpp.h>

#include <iostream>
#include <memory>

class A {
public:
    void open(const char* path);
    void insertValue(const char* value);

private:
[...]

PR #225, opened 2019-10-11
Add default constructor and open() method

kirillin:
Sometimes a separate open method is vital.

### DIFF
--- a/include/SQLiteCpp/Database.h
+++ b/include/SQLiteCpp/Database.h
@@ -76,6 +76,9 @@ class Database
     friend class Statement; // Give Statement constructor access to the mpSQLite Connection Handle
 
 public:
+    // Empty constructor
+    Database();
+
     /**
      * @brief Open the provided database UTF-8 filename.
      *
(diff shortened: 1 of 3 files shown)


SRombauts:
Hello, thanks for this!

Have you noticed CI complains about a lesser code coverage with your change? 
It would be because you don't test all the new code paths. 

I am quite reluctant to have this change; it modifies a lot the behavior/semantic od the Database object...  It needs more thinking, as well as more testing 

Also, your default constructor should really Initialize all variables.

SRombauts:
About the test coverage decrease, you would need to test all cases in the new Database::open() method
(see edit: https://coveralls.io/builds/27254814/source?filename=src/Database.cpp)

But actually, it would be a lot better to have my old constructor call this new open method itself, so you would not need any more unit tests, it would be covered by specific existing ones

SRombauts:
See also #103 dynamically create 'DB' item and use it until close

tzpeng:
//Add close() method
//Fix sqlite3_close sometimes return SQLITE_BUSY
int ret = sqlite3_close(db); 
while (ret == SQLITE_BUSY) {
	ret = SQLITE_OK;
	
	sqlite3_stmt *stmt = sqlite3_next_stmt(db, NULL);
	if (stmt != NULL) {
		ret = sqlite3_finalize(stmt);
		if (ret == SQLITE_OK) {
			ret = sqlite3_close(db);
		}
	}
}

PR #228, opened 2019-11-08
CMakeLists.txt: use transitive compile definitions via cmake

emmenlau:
This PR adds transitive compile definitions via cmake. When this PR is used, a slightly more "modern" syntax is employed for the compile defines `SQLITE_ENABLE_COLUMN_METADATA`, `SQLITECPP_ENABLE_ASSERT_HANDLER`, and `SQLITE_USE_LEGACY_STRUCT`. With this more modern syntax, its possible to specify for every flag a `PUBLIC` or `PRIVATE` visibility. With `PUBLIC` visibility (currently the default in this PR), downstream projects also have access to the flags.

This can be specifically helpful for `SQLITE_ENABLE_COLUMN_METADATA`, which downstream projects can use to tweak their code at compile time.

Please feel free to change the visibility of the individual flags. The previous default (before this PR) was `PRIVATE` visibility for all three flags.


SRombauts:
Thank you, this looks promising, but have you seen that all builds are failing?

```
libSQLiteCpp.a(Column.cpp.o): In function `SQLite::Column::getOriginName() const':
/home/travis/build/SRombauts/SQLiteCpp/src/Column.cpp:51: undefined reference to `sqlite3_column_origin_name'
libSQLiteCpp.a(Statement.cpp.o): In function `SQLite::Statement::getColumnOriginName(int) const':
/home/travis/build/SRombauts/SQLiteCpp/src/Statement.cpp:379: undefined reference to `sqlite3_column_origin_name'
collect2: error: ld returned 1 exit status
make[2]: *** [SQLiteCpp_example1] Error 1
make[1]: *** [CMakeFiles/SQLiteCpp_example1.dir/all] Error 2
```

```
SQLiteCpp.lib(Statement.obj) : error LNK2019: unresolved external symbol _sqlite3_column_origin_name referenced in function "public: char const * __thiscall SQLite::Statement::getColumnOriginName(int)const " (?getColumnOriginName@Statement@SQLite@@QBEPBDH@Z) [C:\projects\sqlitecpp\build\SQLiteCpp_example1.vcxproj]
```

emmenlau:
Oh, thanks, I did not realize! I'll take a look...

emmenlau:
Dear @SRombauts, can you help me shed some light on this? I am under the impression that the `CMakeLists.txt` is correct. The build error we keep on getting indicates that at least the travis build tries to link `sqlite` but the provided `sqlite` does not have support for `_sqlite3_column_origin_name()`. I fail to see how this may have worked before, because option `SQLITE_ENABLE_COLUMN_METADATA` defaulted to `ON` already before.

emmenlau:
Ah I see the issue now.

emmenlau:
I've copied the define also to the supplied `sqlite3/CMakeLists.txt`, sorry I overlooked this before.

Actually, due to the transitiveness of the new cmake flags, it would be sufficient to *only* have this option in `sqlite3/CMakeLists.txt`, as long as users would use the supplied sqlite. But to more more flexible with external sqlite, I think its fine to duplicate this option.

emmenlau:
Now I see more green lights :-)

I hope the PR is fine like this?

SRombauts:
Yes, thanks a lot for this!

emmenlau:
Thanks to you!

PR #230, opened 2019-11-25
Fixed installation on other than Ubuntu GNU/Linux distributions

xvitaly:
Fixed installation on other than Ubuntu GNU/Linux distributions.

Fedora and other GNU/Linux distributions use different $libdir prefixes. Now it can be installed on any GNU/Linux distributions.


SRombauts:
Thanks for your contribution!

PR #231, opened 2019-11-25
Added SOVERSION field for shared library

xvitaly:
Most of GNU/Linux distributions require SOVERSION field of shared library to be set.

You should bump SOVERSION every time you draft a new release with breaking API changes.


SRombauts:
Thank you :)

PR #232, opened 2019-11-25
Added support of packaged GTest for running unit tests

xvitaly:
Added support of packaged GTest for running unit tests with fallback to bundled version if not found.


SRombauts:
Thanks again!

PR #235, opened 2019-12-14
Added support for cmake dependencies and version information

emmenlau:
This PR adds support for cmake transitive dependency tracking and cmake version information. With this PR, when SQLiteCpp links against an external SQLite3, then downstream projects will also automatically link against SQLite3.

Also, this PR enables using the version number in the find_package command:
```
find_package(SQLiteCpp 2.4.0)
```


SRombauts:
Thanks for this!

PR #237, opened 2019-12-14
Allow optional c++11 in the build, and require c++11 for tests

emmenlau:
With Google test 1.10, c++11 is mandatory (see https://github.com/google/googletest/issues/1366#issue-283686739). To build the tests, this PR requires c++11 support. Additionally, the PR adds a new build flag that enables c++11 for the whole library. By setting `SQLITECPP_ENABLE_CPP11` to `TRUE` the library will use c++11 features.


SRombauts:
Hello @emmenlau thank you for this.

I really like this, and modern C++ in general, but for now I have maintained explicit support for pre-C++11, so CI is explicitly testing with C++98 (and fails with this).

I you want to help moving forward, you could disable these CI config using C++98, force enable C++11 or above, add a note to the readme, and increase the lib version accordingly

Cheers

SRombauts:
Closing this one since C++11 is now up an running

PR #239, opened 2019-12-24
Enabled C++ 20 if it is available.

KOLANICH:


KOLANICH:
>CXX_STANDARD is set to invalid value '20'

Old versions of CMake should be dropped.

KOLANICH:
I have no idea why https://travis-ci.org/SRombauts/SQLiteCpp/jobs/629439562 fails

KOLANICH:
????

SRombauts:
Ah ah, sorry, I merged the branch 3.x into master and deleted it an hour ago, which seems to to have close the PR automatically: please retarget it to master.

KOLANICH:
@SRomnauts

SRombauts:
Hello, so in this case, I don't think it would be a good idea to have a library try to use C++ latest standard.
I would rather have a minimum/required version (C++11 right now) and the application using it deciding what is best.

What would be interesting is to add more test cases to the Github Actions to test with latest compilers and latest C++ standard in CI

KOLANICH:
1. It is optional. I mean it has to be explicitly enabled. Of course I can provide the needed CMake option via CLI, butnit has some logic dependent on CMake version.
2. If you still dislike it, feel free to close the PR

CMakeLists.txt
```diff
@@ -10,7 +10,23 @@ project(SQLiteCpp VERSION "2.99")
 
 # SQLiteC++ 3.x now requires C++11 compiler
 set(CMAKE_CXX_STANDARD 11)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+option(USE_LATEST_CXX_STANDARD_VERSION "Makes CMake to probe for the latest CXX version. May give efficiency improvements. Disables mandating minimal standard version: CMake has design flaws that prevents from adding a minimal version constraint while opportunistically upgrading: CMAKE_CXX_STANDARD_REQUIRED should have been an int for that, not a boolean" OFF)
```
SRombauts: Could you please rename it to USE_LATEST_CXX_STANDARD and make the description shorter by moving the note about CMake design flaw to a simple comment above the option

CMakeLists.txt
```diff
@@ -10,7 +10,23 @@ project(SQLiteCpp VERSION "2.99")
...
+        set(CMAKE_CXX_STANDARD 20)
+    else()
+        if(${CMAKE_VERSION} VERSION_GREATER "3.8") 
+            set(CMAKE_CXX_STANDARD 17)
+        else()
+            set(CMAKE_CXX_STANDARD 14)
+        endif()
+    endif()
+else()
+    set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
SRombauts: could you also move the line set(CMAKE_CXX_STANDARD 11) here so it will make more sense (to have it all in the same if/else)

CMakeLists.txt
SRombauts: this one is good now :)

emmenlau:
@KOLANICH would https://github.com/SRombauts/SQLiteCpp/pull/269 also solve your needs? I think it requires less changes but gives you the same flexibility. Is that good?

SRombauts:
dropped in favor of #269

PR #240, opened 2019-12-24
Replaced the bundled sqlite sources with the oes retrieving the lib from the official website.

KOLANICH:


KOLANICH:
>CMake 3.14 or higher is required. You are running version 3.9.2

Don't remember why I have required this version, probably something FetchContent-related

KOLANICH:
????

PR #241, opened 2019-12-24
Improved efficiency and code reuse with templates and pure functions

KOLANICH:


SRombauts:
Hello, thanks again for this new series of pull requests. Have you seen that they have all failed CI?

KOLANICH:
>Hello, thanks again for this new series of pull requests.

Not quite new, just splitted the old #182 into multiple PRs.

>no template named 'enable_if' in namespace 'std'
>error: ‘enable_if’ in namespace ‘std’ does not name a template type

Old versions of C++ should be dropped.

emmenlau:
@KOLANICH I find your PR very valuable but I think it may be easier to digest if split into again smaller parts.

@SRombauts What is your take on the points raised by KOLANICH? Lets take one aspect out, templates. Would you agree that adding stronger use of templates is a good thing? I recently had to fix something about the types in `Statement::bind()`, and I found what KOLANICH also found, that these methods are quite lengthy code for a thing that would become much shorter and cleaner with templates.

Would you accept a PR that adds stronger use of templates (for better code reuse and shortening of the existing code)?

SRombauts:
I am closing this as outdated, but we can reopen the discussion on the basis of smaller scoped PRs

PR #242, opened 2019-12-25
Efficiency splitted

KOLANICH:
Splitted #241 even further. Should be a bit more reviewable what I have done than when collapsed into a single commit. Though it is not yet fully usable, let's try to troubleshhot step-by-step.


KOLANICH:
@SRombauts, Travis passes

SRombauts:
Hello @KOLANICH thank you very much for splitting your changes!

This one is very good indeed: I was actually starting to work on a similar refactor this morning when I was remembered of it.

After this get in, please don't hesitate to rebase/retarget your other pull request so that we can have some more in. Basically I am only looking at a PR seriously when it passes all checks (which can lead to some wasted like this one).

Cheers!

KOLANICH:
Thanks for merging.

PR #245, opened 2019-12-26
SQLiteCpp 3.x more modern C++11 & CMake 3.1

SRombauts:
Hello all, hello @KOLANICH @emmenlau @xvitaly 

I am starting a new branch for a cleaner version of this wrapper.

I think we could aim for C++14 and a more recent CMake version (this also depends on Travis CI)

The goal is to modernize a bit the code, but more important to me is to try to remove macros & #ifdef since it's not good for readability & maintainability.

Cheers,
Sébastien


KOLANICH:
>this also depends on Travis CI

IMHO we shouln't use Travis. We should use GH Actions and GL CI

SRombauts:
Oh yes, I should give it a look now that it is out!

prsolucoes:
Hi,

@SRombauts please don't put c++14 as a requirements. I use it in a lot of projects with c++11.

Thanks.

SRombauts:
> @SRombauts please don't put c++14 as a requirement. I use it in a lot of projects with c++11.
Ok, it's always good to know. I think C++11 is a perfectly good target

For now, my goal would be to test oldest versions with Ubuntu 14.04 with GCC 4.8.4 / Clang 5 / CMake 3.9

emmenlau:
Sounds all great, I'm happy to test and help out after the holidays...

SRombauts:
@prsolucoes is it okay that I remove support for Visual Studio 2013?
It was only partial with a bunch of warnings. It is not able to build Googletest 1.10 either.

Visual Studio 2015, 2017 & 2019 are all much more modern & mature, with incremental updates, and better support for C++ standards

prsolucoes:
@SRombauts to me no problem. I use only 2017. Thank you for your concern.

PR #249, opened 2019-12-30
Added SQLite header parsing functionality and associated tests

patrick--:
Added header parsing functionality via a `getHeaderInfo()` function. This function reads the first 100 bytes of a SQLite database file and attempts to reconstruct byte groups into the specific header fields within a Header object. Also added tests to demonstrate updating header values via PRAGMA and verifying updated / default header values.

### DIFF
--- a/include/SQLiteCpp/Database.h
+++ b/include/SQLiteCpp/Database.h
@@ -12,7 +12,7 @@
 
 #include <SQLiteCpp/Column.h>
 #include <SQLiteCpp/Utils.h>    // definition of nullptr for C++98/C++03 compilers
-
+#include <cstdint>
 #include <string.h>
 
 // Forward declarations to avoid inclusion of <sqlite3.h> in a header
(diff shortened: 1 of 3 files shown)


patrick--:
@SRombauts tell me what you think of this implementation. In the meantime I am looking into getting the tests to pass but still have access to fixed width integer datatypes.

SRombauts:
About the compile error, you can use plain old C type (unsigned char/short/int). Alternatively you could apply your pull request to the slqitcpp-3.x branch that is c++11

patrick--:
@SRombauts I am going to go ahead and replace with plain C types for now so it can be available in master. I'll also open a PR with fixed with support for the C+11 branch. Thanks!

SRombauts:
I left a bunch of comments, but it looks good overall!

SRombauts:
Also don't bother to create a different version for c++11, I don't want that we go an rewrite all just for the sake of it

SRombauts:
And finally, I have just reached 100% unit test code coverage (for what it's worth) so you should test all cases of error

https://coveralls.io/builds/27840818/source?filename=src/Database.cpp

Basically you have to test also with no filename, no file, and wrong file hezder

patrick--:
@SRombauts okay I will get those tests added.

patrick--:
@SRombauts you mentioned leaving some comments? I might be blind, but I'm not seeing those anywhere.

SRombauts:
Sorry, my bad, I didn't know this new review system needed a submit when finished

SRombauts:
That's very good, thanks a lot!

PR #261, opened 2020-01-14
Replaced the bundled sqlite sources with the ones retrieving the lib from the official website.

KOLANICH:


KOLANICH:
@SRombauts

SRombauts:
Hello @KOLANICH
I actually don't think it's a good idea to not have a copy of SQLite3 source code directly on the.
I also cannot request all users to move to such a recent CMake 3.14, it's not supported by popular CI systems

Perhaps could you make the Fetch feature optional?

Cheers!

SRombauts:
Same, let's redo the work in a targeted manner if it's something still interersting

PR #263, opened 2020-01-17
Use modern cmake syntax for c++11 option

emmenlau:
The current cmake implementation uses old-style cmake syntax with global variables to set C++11 and Position Independent Code properties. This PR changes the syntax to modern cmake style.

Is this good?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,10 +8,6 @@ cmake_minimum_required(VERSION 3.1) # for "CMAKE_CXX_STANDARD" version
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake") # custom CMake modules like FindSQLiteCpp
 project(SQLiteCpp VERSION "2.99")
 
-# SQLiteC++ 3.x now requires C++11 compiler
-set(CMAKE_CXX_STANDARD 11)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-
 message (STATUS "CMake version: ${CMAKE_VERSION}")
 message (STATUS "Project version: ${PROJECT_VERSION}")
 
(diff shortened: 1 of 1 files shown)


SRombauts:
I am unsure of the benefit of this.

What do you think of this other PR #239 that will be incompatible with this?
Basically, we should enforce a minimum version, not an explicit version.
So if people want to use C++17 it's fine, while right now I may have enforced C++11 for everyone (losing some feature we have in the wrapper)

emmenlau:
I fully understand. But it may help to split the "question" into two parts:

I do recommend to use modern cmake syntax wherever possible. This is because it brings often underlying benefits, like transitivity of properties, and easier maintenance of the code in the future. For example, setting position-independent code can be passed down via cmake flags to consumers of this library. My PR does not (yet) include this change. But it makes better use of modern cmake syntax than the current code. So my PR is only about the syntax.

I do *not* mean to suggest that a higher c++ standard should be enforced. I agree that this should be up to the user of the library. While I think that c++11 is fine now, I still agree that more modern features may not bring a big benefit to the larger community but may exclude many users. My personal recommendation would be to make cmake options for specific features that require a higher c++ standard. For example add something like:
```
option(ENABLE_XXX "Allow to use feature XXX. Requires c++14." OFF)
if(ENABLE_XXX)
    target_compile_definitions(${PROJECT_NAME} PUBLIC SQLITECPP_ENABLE_XXX)
endif()
```

Users will be able to choose for the features they require, and you can deduce the required c++ standard based on that. Personally I think that is the gold standard you can achieve. But it makes your code sightly more complex because you need to separate out modern features based on the additional defines.

What do you think?

emmenlau:
Oh and by the way, I think PR #239 and my change here are perfectly compatible. Only the syntax of #239 would need to be slightly changed.

emmenlau:
Closing this in favor of #269

PR #269, opened 2020-01-30
CMakeLists.txt: Make c++11 optional unless building tests

emmenlau:
Several minor changes with respect to c++ standard:
  * Changing `CMakeLists.txt` to make c++11 optional. This is a good idea because SQLiteCpp can support even older compilers. With this change, c++11 becomes requested but the build can proceed if it is not found.
  * When tests are enabled, c++11 becomes mandatory. The user is informed about this setting via cmake output.
  * Allow users to specify other c++ versions via the cmake options: `-DCMAKE_CXX_STANDARD="[98, 11, 14, 17, 20]"`

Its my understanding that this PR is more generic than https://github.com/SRombauts/SQLiteCpp/pull/239 because it allows users to manually specify a newer c++ standard, without all the special logic and extra options.

This PR may also supersede https://github.com/SRombauts/SQLiteCpp/pull/263 since it is more generic than the other PR.

Comments and feedback welcome.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,8 +8,13 @@ cmake_minimum_required(VERSION 3.1) # for "CMAKE_CXX_STANDARD" version
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake") # custom CMake modules like FindSQLiteCpp
 project(SQLiteCpp VERSION 2.99)
 
-# SQLiteC++ 3.x now requires C++11 compiler
-set(CMAKE_CXX_STANDARD 11)
+# SQLiteC++ 3.x requires C++11 features
+if (NOT CMAKE_CXX_STANDARD)
+    set(CMAKE_CXX_STANDARD 11)
+elseif (CMAKE_CXX_STANDARD LESS 11)
+    message(WARNING "CMAKE_CXX_STANDARD has been set to '${CMAKE_CXX_STANDARD}' which is lower than the minimum required standard (c++11).")
+endif ()
+message(STATUS "Using c++ standard c++${CMAKE_CXX_STANDARD}")
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 message (STATUS "CMake version: ${CMAKE_VERSION}")


SRombauts:
Thanks, I'll look into it seriously right away.

There is a change needed though: upcoming SQLiteCpp 3.x definitely requires C++11 and will fail with older compilers.
I have started migrating some code to std::unique_ptr<> with use of move constructors & operatos

CMakeLists.txt
```diff
@@ -8,9 +8,18 @@ cmake_minimum_required(VERSION 3.1) # for "CMAKE_CXX_STANDARD" version
...
 project(SQLiteCpp VERSION 2.99)
 
-# SQLiteC++ 3.x now requires C++11 compiler
-set(CMAKE_CXX_STANDARD 11)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
+# SQLiteC++ 3.x can make use of C++11 features
+if (NOT CMAKE_CXX_STANDARD)
+    set(CMAKE_CXX_STANDARD 11)
+endif ()
+# googletest requires C++11 features
```
SRombauts: SQLiteCpp 3.x on master branch already requires C++11

CMakeLists.txt
```diff
@@ -8,9 +8,18 @@ cmake_minimum_required(VERSION 3.1) # for "CMAKE_CXX_STANDARD" version
...
+if (SQLITECPP_BUILD_TESTS)
+    if (CMAKE_CXX_STANDARD LESS 11)
+        message(STATUS "googletest requires c++ standard 11 or higher")
+    endif ()
+    message(STATUS "Setting c++ standard ${CMAKE_CXX_STANDARD} to required for unit tests with googletest")
+    set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
SRombauts: I don't really see the point of (CMAKE_CXX_STANDARD_REQUIRED ON) since my understanding is that it will enforce C++11 usage even if the user has selected C++17

=> isn't it possible to use a fatal error instead of the above message

actually, I am not sure of the behavior of the CMAKE_CXX_STANDARD variable if we try to set it to a higher version than supported?

CMakeLists.txt
emmenlau: Ok, then I will make `CMAKE_CXX_STANDARD_REQUIRED ON` unconditional, and I can add a warning if users set a lower standard than `11`.

CMakeLists.txt
emmenlau: The point of `CMAKE_CXX_STANDARD_REQUIRED ON` was that I assumed you wanted to still support older compilers, and without this setting `c++11` was recommended but not required.

But no worries about newer standards. When setting `CMAKE_CXX_STANDARD_REQUIRED ON` it just means "enforce whatever standard is requested in `CMAKE_CXX_STANDARD`, which can still be set to any legal version".

So there is no problem for users to still select `-DCMAKE_CXX_STANDARD=17` or `-DCMAKE_CXX_STANDARD=20`.

emmenlau:
Dear @SRombauts , the newest PR should respect your suggestions. With these changes, the minimum c++ standard will be 11, but other versions are allowed. If the user selects a *lower* standard than 11 they will get a warning. c++11 or higher standards can be selected unconditionally, and will always be treated as `required`, so cmake should fail if the compiler does not support it.

SRombauts:
Thanks a lot, I like it!

Issue #273, opened 2020-02-11
CMake files not installed in same lib directory


tcraigtyler:
SQLiteCpp 3.0.0, RHEL 7.7 x86_64, CMake 3.14.6

`make install` creates cmake files in both lib/cmake and lib64/cmake directories:
```
-- Installing: /usr/local/lib64/cmake/SQLiteCpp/SQLiteCppTargets.cmake
-- Installing: /usr/local/lib64/cmake/SQLiteCpp/SQLiteCppTargets-noconfig.cmake
-- Installing: /usr/local/lib/cmake/SQLiteCpp/SQLiteCppConfig.cmake
-- Installing: /usr/local/lib/cmake/SQLiteCpp/SQLiteCppConfigVersion.cmake
```
Trying to use find_package(SQLiteCpp) then results in the following error message:
```
-- Found SQLite3: /usr/local/include (found version "3.30.1") 
CMake Error at /usr/local/lib/cmake/SQLiteCpp/SQLiteCppConfig.cmake:30 (include):
  include could not find load file:

    /usr/local/lib/cmake/SQLiteCpp/SQLiteCppTargets.cmake
```

PR #274, opened 2020-02-11
Install cmake files into same directory

tcraigtyler:
Addresses #273


SRombauts:
Good catch, thank you for your contribution!

Issue #276, opened 2020-04-05
CMake fails on Ubuntu


travishaagen:
Using Ubuntu 19 (on arm64), I have no trouble compiling SQLiteCpp, but my own CMake configuration fails with the following error:

```
CMake Error at /usr/share/cmake-3.13/Modules/CMakeFindDependencyMacro.cmake:48 (find_package):
  By not providing "FindSQLite3.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "SQLite3", but
  CMake did not find one.

  Could not find a package configuration file provided by "SQLite3" with any
  of the following names:

    SQLite3Config.cmake
    sqlite3-config.cmake

  Add the installation prefix of "SQLite3" to CMAKE_PREFIX_PATH or set
  "SQLite3_DIR" to a directory containing one of the above files.  If
  "SQLite3" provides a separate development package or SDK, be sure it has
  been installed.
Call Stack (most recent call first):
  /usr/local/lib/cmake/SQLiteCpp/SQLiteCppConfig.cmake:2 (find_dependency)
  CMakeLists.txt:50 (find_package)
```

I'm guessing that since `SQLiteCppConfig.cmake` contains the line `find_dependency(SQLite3)` that it is telling CMake that either `SQLite3Config.cmake` or `sqlite3-config.cmake` exist on system. However, it looks like this Ubuntu issue is stating that these files are not currently installed:

https://bugs.launchpad.net/ubuntu/+source/sqlite3/+bug/1868166

I had thought that when `cmake -DSQLITECPP_INTERNAL_SQLITE=ON ..` is used, that it might make everything work for me, so perhaps this config option could be checked for in https://github.com/SRombauts/SQLiteCpp/blob/master/cmake/SQLiteCppConfig.cmake.in to stop the execution of `find_dependency(SQLite3)`.

travishaagen:
On another note, I have been compiling on Mac and Alpine Linux prior to trying Ubuntu and ran into other CMake issues with SQLite3. I found the following that seemed to work:

```
find_library(SQLITE3_LIB libsqlite3.a sqlite3)
if (NOT SQLITE3_LIB)
    message(FATAL_ERROR "sqlite3 library not found")
endif ()
```

SRombauts:
Oh yes I see, I think that SQLiteCppConfig doesn't work with/has not been thought for use with the Internal SQLite3

[...]

PR #277, opened 2020-05-08
Fix cmake scoping issues with packaged SQLite

peterbell10:
Linking against the packaged version of `sqlite3` only added include paths when installed, not during build, due to being included as `PRIVATE`. I assume this was done because `SQLiteCpp` users shouldn't need to include `sqlite.h`, so I've  also changed `SQLiteCpp` to link privately to `sqlite`.

Now `sqlite` correctly propagates its include directory when linked, but it is not propagated to users linking against `SQLiteCpp`. These changes are useful because internal and external sqlite now behave the same.


sum01:
I believe the `find_dependency(SQLite3 REQUIRED)` call in `cmake/SQLiteCppConfig.cmake.in` should also be removed when doing this.

peterbell10:
IIUC, I think it's still required. Private dependencies still need to be linked transitively when built as shared libs. So, the installed cmake still depends on sqlite3.

SRombauts:
Hello, thank you for this proposal. 

I think it's actually thé opposite that we want: users sometimes need to include sqlite.h themselves when they want access to any C function that is not wrapped here

peterbell10:
Okay, I was just going for consistency with how it worked already. I've reversed the private linking.

SRombauts:
Yes, I understand, no worries :)

This CMake code has changed a lot recently, and a big part of its features are not tested.

I am a strong believer that only the automatically tested features can be properly maintained!
So I'll try to implement a few basic examples of usage of SQLiteCpp

nidefawl:
Any progress on this?
Using SQLiteCpp with find_package has been broken for almost 2 years now.
See https://github.com/SRombauts/SQLiteCpp/issues/276

SRombauts:
Hey @peterbell10, I am so sorry, I intended to finally try to merge this one and test everything but it is no in conflict

SRombauts:
Hey, thanks a lot for providing this; I am now going to finally give it a go!

PR #278, opened 2020-05-14
Add missing Threads dependency to Config.cmake.in

sum01:
Without this, systems that linked against Threads::Threads (namely
Linux) don't "just work" with find_package(SQLiteCpp), but instead
require you to find & link against Threads manually.

So this fixes that by finiding it for you, if it was used.


SRombauts:
Thanks !

PR #281, opened 2020-05-31
Added cmake variables to make submoduling the project easier

abelmul:
That explains it


SRombauts:
Hello, thanks for this proposal.

I think that what you should do here is use the provided Find SQLiteCpp cmake file instead, no ?

Cheers

abelmul:
To quote the readme file in order to submodule and add this library you have to do

```cmake
add_subdirectory(${CMAKE_CURRENT_LIST_DIR}/thirdparty/SQLiteCpp)

include_directories(
  ${CMAKE_CURRENT_LIST_DIR}/thirdparty/SQLiteCpp/include
)

add_executable(main src/main.cpp)
target_link_libraries(main
  SQLiteCpp
  sqlite3
  pthread
  dl
  )
```
 With this pull request you can do
``` cmake
add_subdirectory(${CMAKE_CURRENT_LIST_DIR}/thirdparty/SQLiteCpp)

include_directories(${SQLITECPP_INCLUDE_DIRS})
add_executable(main src/main.cpp)
target_link_libraries(main ${SQLITECPP_LIBRARIES)
```

I think it keeps application code clean.

ChrisThrasher:
If linking multiple libraries is required then that hints at the fact that the SQLiteCpp target is missing some `PUBLIC` dependencies. That's the modern CMake solution and is preferable to the old school approach of having `LIBNAME_INCLUDE_DIRS` and `LIBNAME_LIBRARIES` variables.

SRombauts:
I am going to close this, as the comment is correct: *_INCLUDE_DIRS / *_LIBRARIES variables are old-style CMake; transitive target dependencies are the modern solution.

PR #287, opened 2020-07-21
Fixed installation on other than Ubuntu distributions

xvitaly:
Fixed installation on other than Ubuntu distributions.


SRombauts:
Thanks for taking the time to report this and provide the fix!

PR #288, opened 2020-07-29
Allow building of sqlite JSON1 extension when building internal sqlite library

zxey:
I added cmake option to enable json1 extension when building internal sqlite library.
I added the option under the SQLITECPP_INTERNAL_SQLITE `if` statement so it is clear that this option is only valid when building the internal sqlite library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -228,6 +228,7 @@ endif ()
 option(SQLITECPP_INTERNAL_SQLITE "Add the internal SQLite3 source to the project." ON)
 if (SQLITECPP_INTERNAL_SQLITE)
     message(STATUS "Compile sqlite3 from source in subdirectory")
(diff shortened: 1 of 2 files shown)


SRombauts:
Hi @zxey I wasn't aware of this extension, do you know since when it exists?

Anyway, I think it's a great feature 👍 

The only small change I would propose is to actually set SQLITE_ENABLE_JSON1 publicly so the client code would be able to test for this feature.

zxey:
It seems that it has been available since sqlite version 3.9.0 as seen in changelog https://sqlite.org/search?s=c&q=JSON1

SRombauts:
See PR https://github.com/SRombauts/SQLiteCpp/pull/428 for the new branch I just pushed:
I am going to remove option SQLITE_ENABLE_JSON1 since the JSON functions and operators are built into SQLite by default, as of SQLite version 3.38.0 (2022-02-22).

For us with SQLiteCpp it would be since Version 3.2.0 - 2022 September 18

PR #292, opened 2020-08-11
Fix compilation if using SQLITE_HAS_CODEC

sum01:
It was failing to compile since it wasn't linking against the sqlcipher
library if using the system libraries. This fix simply links the library
& includes the header path.

PkgConf is optional in this process, and it will fall back to regular
search for header & lib if it's not installed.


SRombauts:
Good job, thank you!

PR #293, opened 2020-08-18
Remove FindSQLiteCpp.cmake

### LINKED ISSUE #283: Remove obsolete FindSQLiteCpp.cmake
Why maintain a FindSQLiteCpp.cmake file when the export is already generated & installed via Cmake?
Exported XXXConfig.cmake files serve the same purpose.

If someone can provide a PR it would be awesome, with perhaps a note on the README for how to use the Config file

sum01:
It's not needed since the automatically generated file
SQLiteCppConfig.cmake already exists.

Added a small section to the README to explain how it can be used when
this library is installed to the system.

Fixes #283


SRombauts:
Hello, thanks for your work!
Have you seen its failing ci builds?
It's just because the file is still used in the CMakeLists.txt

    Cannot find source file:
        cmake/FindSQLiteCpp.cmake

sum01:
Should be all good. Just forgot to remove that line like you said.

SRombauts:
Thanks again!

PR #295, opened 2020-08-23
Compile internal SQLite library with -ffunction-sections

smichaku:
When building with SQLITECPP_INTERNAL_SQLITE=ON the SQLite amalgamation
source is used for generating the library. Using one big source file
means all the library code will be put in a single section. When
building statically linked executables the entire section will be
linked even if a small portion of the library is actually used.

This commit addresses this issue by setting the -ffunction-sections
compiler option when building the library. As each function is placed in
a section of its own the linker, when passed the --gc-sections, will
throw away unused sections (functions) and reduce the executable size.


SRombauts:
Thanks a lot for providing this!
I am sorry that it took me so long to merge it, especially since I did review it early, but then it got out of my mind.

PR #305, opened 2020-10-08
Add other constants that work with sqlite3_open_v2

LuAPi:
Add the other flags used in the open function to the SQLite namespace for convenience.

I added the ones that are ok to use with the `sqlite3_open_v2` function described here https://www.sqlite.org/c3ref/open.html .
I didn't add the VFS only flags mentioned here https://www.sqlite.org/c3ref/c_open_autoproxy.html .

I also made sure the comments relating to threading mode used the "multi-thread" and "serialized" terminology used on this page https://www.sqlite.org/threadsafe.html for clarity.

I'm not sure if there was a reason these weren't included but the first thing I wanted to do was open the database in multi-thread mode and I found the flag was missing.

I've not really tested this beyond compiling it and opening a database with the `SQLite::OPEN_NOMUTEX` flag, but since this is just declaring some consts I expect the compiler is enough to check it's right.


SRombauts:
Thanks for your contribution, I don't know exactly how I didn't merge it when I first saw it, but now it's fixed 👍

Issue #306, opened 2020-10-11
MinGW compilation warning in pedantic mode


chronoxor:
MinGW generates warning with gcc 8.1.0 in pedantic mode:
```
Z:\projects\Projects\PrimeTrader\modules\SQLiteCpp\src\Database.cpp: In static member function 'static SQLite::Header SQLite::Database::getHeaderInfo(const string&)':
Z:\projects\Projects\PrimeTrader\modules\SQLiteCpp\src\Database.cpp:313:12: warning: 'char* strncpy(char*, const char*, size_t)' output may be truncated copying 16 bytes from a string of length 99 [-Wstringop-truncation]
     strncpy(pHeaderStr, pBuf, 16);
     ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
```

SRombauts:
Fixed by linked merge request #308

PR #307, opened 2020-10-11
MinGW compilation warning in pedantic mode #306

chronoxor:


stauffer-garmin:
@chronoxor If you're interested, I proposed a fix for this in #308 which doesn't change the build flags.

SRombauts:
Thanks, but I have merged the fix for string truncation instead.

PR #308, opened 2020-11-09
Fix build warning due to string truncation

stauffer-garmin:
strncpy gives an "output may be truncated" warning in newer versions of
GCC due to *pBuf being larger (100) than *pHeaderStr (16).  Use memcpy
and explicitly null-terminate the target string.

This fixes issue #306.


SRombauts:
Looking good, thank you!

PR #311, opened 2020-11-12
Add Database::tryExec()

kcowolf:
Similar to Statement::tryExecuteStep(), allows calls to sqlite3_exec() with the caller assuming responsibility for checking and handling the returned status.

### DIFF
--- a/include/SQLiteCpp/Database.h
+++ b/include/SQLiteCpp/Database.h
@@ -197,6 +197,7 @@ public:
      *  - Data Definition Language (DDL) statements "CREATE", "ALTER" and "DROP"
      *  - Data Control Language (DCL) statements "GRANT", "REVOKE", "COMMIT" and "ROLLBACK"
      *
+     * @see Database::tryExec() to execute, returning the sqlite result code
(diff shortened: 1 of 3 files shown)


SRombauts:
Hello!
I find this to be useful. But I am really hesitant about the name, since it doesn't convey exactly the fact that it doesn't deal with exceptions (quite the opposite with the "try" keyword)
Or is this a sort of naming convention I am unaware of?

kcowolf:
I was trying to follow Statement::tryExecuteStep(), which similarly doesn't throw exceptions (compared to Statement::executeStep and Statement::exec, which do).

Also, C# has some "Try" functions which behave similarly.  For example, int.TryParse() versus int.Parse().  int.TryParse() returns false if the string can't be parsed as an integer, while int.Parse() throws an exception.  https://stackoverflow.com/questions/467613/parse-v-tryparse

I'd be fine with renaming it if you prefer, though.  It just made sense to me based on the above.

SRombauts:
Then it's settled, thank you!

PR #313, opened 2020-12-11
[CMake] Add SQLITECPP_INCLUDE_SCRIPT option

past-due:
Add a `SQLITECPP_INCLUDE_SCRIPT` option to support disabling the addition of config & script files to the library.
(The default is `ON`, replicating current behavior.)

Adds flexibility without requiring library users to modify the `CMakeLists.txt` directly.
Also, this is needed for a dependent program that automatically packages the library source (as a dependency) in a source tarball "snapshot" and wants to exclude files not needed for building via CMake locally (.git* and such) to save space and reduce the overall file count.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -153,24 +153,27 @@ set(SQLITECPP_DOC
 )
 source_group(doc FILES ${SQLITECPP_DOC})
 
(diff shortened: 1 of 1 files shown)


SRombauts:
Thank you for this useful contribution :)

Issue #315, opened 2021-01-05
Bug when compile with sqlite 3.34.0


paulocoutinhox:
Hi,

Im getting the following error:

```
[ 85%] Building CXX object source_subfolder/CMakeFiles/SQLiteCpp.dir/src/Transaction.cpp.o
/Users/paulo/.conan/data/sqlitecpp/3.1.1/_/_/build/518d5483cdb53cc130fd28cdad883cd2b4d0a54f/source_subfolder/src/Database.cpp:213:11: error: use of undeclared identifier
      'sqlite3_load_extension'
    ret = sqlite3_load_extension(getHandle(), apExtensionName, apEntryPointName, 0);
          ^
1 error generated.
```

Can you help me?

SRombauts:
Hi! 
Thanks for reporting, though I don't see how that would be possible.
Nothing is ever removed in SQLite https://sqlite.org/changes.html

I suspect that you might be trying to link against à pre-built SQLite version were the extension mechanism is disabled. 

Let me know

paulo-coutinho:
I revert it and i'm using version 2.5.0 and this problem don't happen.

paulo-coutinho:
It only happen for iOS, because on other platform it don't happen.

If i change from:
```
self.requires("sqlitecpp/2.5.0")
```

to:
```
self.requires("sqlitecpp/3.1.1")
```

The error happen.

paulo-coutinho:
I checked now that it happen with sqlitecpp/2.5.0 too:

```
[ 14%] Building CXX object source_subfolder/CMakeFiles/SQLiteCpp.dir/src/Backup.cpp.o
[ 57%] Building CXX object source_subfolder/CMakeFiles/SQLiteCpp.dir/src/Column.cpp.o
[ 57%] Building CXX object source_subfolder/CMakeFiles/SQLiteCpp.dir/src/Database.cpp.o
[ 57%] Building CXX object source_subfolder/CMakeFiles/SQLiteCpp.dir/src/Exception.cpp.o
[ 71%] Building CXX object source_subfolder/CMakeFiles/SQLiteCpp.dir/src/Statement.cpp.o
[ 85%] Building CXX object source_subfolder/CMakeFiles/SQLiteCpp.dir/src/Transaction.cpp.o
[ 85%] Built target SQLiteCpp_cppcheck
[...]

PR #316, opened 2021-01-06
Update SQLite3 from 3.33.0 to 3.34.0 (2020-12-01)

SRombauts:


SRombauts:
In link with #315 issue linking sqlite3_load_extension with SQLite 3.34.0

Issue #317, opened 2021-01-06
Visual Studio 2019 compilation error ('SQLite::Database::Database': delegating constructor calls itself)


chronoxor:
```
SQLiteCpp\include\SQLiteCpp/Database.h(182,17): error C7535: 'SQLite::Database::Database': delegating constructor calls itself
```

chronoxor:
![image](https://user-images.githubusercontent.com/4397805/103817962-b2250400-5078-11eb-8c91-e7561cf5fdf4.png)

chronoxor:
Send a PR with a fix

SRombauts:
Thanks for reporting!
I'll see what to do... 

I'll also have to check why it is not catched by continuous integration (Travis CI and / or Github Actions)

chronoxor:
Filesystem is also issues with MinGW-w64:
```
[ 22%] Building CXX object modules/SQLiteCpp/CMakeFiles/SQLiteCpp.dir/src/Database.cpp.obj
In file included from C:/mingw64/lib/gcc/x86_64-w64-mingw32/8.1.0/include/c++/filesystem:37,
                 from C:/projects/primetrader/modules/SQLiteCpp/include/SQLiteCpp/Database.h:16,
                 from C:/projects/primetrader/modules/SQLiteCpp/include/SQLiteCpp/Backup.h:14,
                 from C:\projects\primetrader\modules\SQLiteCpp\src\Backup.cpp:12:
C:/mingw64/lib/gcc/x86_64-w64-mingw32/8.1.0/include/c++/bits/fs_path.h: In member function 'std::filesystem::__cxx11::path& std::filesystem::__cxx11::path::operator/=(const std::filesystem::__cxx11::path&)':
C:/mingw64/lib/gcc/x86_64-w64-mingw32/8.1.0/include/c++/bits/fs_path.h:237:47: error: no match for 'operator!=' (operand types are 'std::filesystem::__cxx11::path' and 'std::filesystem::__cxx11::path')
    || (__p.has_root_name() && __p.root_name() != root_name()))
                               ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~

```

Chronial:
@SRombauts The default cmake build does not build with c++17, so the problematic code is disabled via `#ifdef`: https://github.com/SRombauts/SQLiteCpp/blob/master/CMakeLists.txt#L11

PR #318, opened 2021-01-06
Visual Studio 2019 compilation error ('SQLite::Database::Database': delegating constructor calls itself) #317

chronoxor:


SRombauts:
Hi, thanks but it means you are disabling this new constructor for vs2017 as well...
Would be better to have a proper fix

Chronial:
I created an alternative fix in #319

PR #319, opened 2021-01-08
Fix std::filesystem::path Database constructor (fixes #317)

### LINKED ISSUE #317: Visual Studio 2019 compilation error ('SQLite::Database::Database': delegating constructor calls itself)

Chronial:
The return type of `path.c_str()` is platform-dependent and is a `wstring` on windows. `path.u8string()` always returns an utf8-encoded string.

In c++17, that string has type `std::string`, but starting in c++20 it is `std::u8string`. I used a reinterpret_cast to handle the c++20 case.

Alternatively, a constructor taking u8string could be added for the c++20 case. But it seems odd to have only one place where the library accepts u8strings. If support for u8strings is added, it should probably be done for the whole library at once.

### DIFF
--- a/include/SQLiteCpp/Database.h
+++ b/include/SQLiteCpp/Database.h
@@ -179,7 +179,8 @@ public:
              const int   aFlags         = SQLite::OPEN_READONLY,
              const int   aBusyTimeoutMs = 0,
              const std::string& aVfs            = "") :
-        Database(apFilename.c_str(), aFlags, aBusyTimeoutMs, aVfs.empty() ? nullptr : aVfs.c_str())
+        Database(reinterpret_cast<const char*>(apFilename.u8string().c_str()),
+                 aFlags, aBusyTimeoutMs, aVfs.empty() ? nullptr : aVfs.c_str())
     {
     }
 


chronoxor:
I confirm that the fix works well for VS2019, but not for MinGW64 8.1.0 due to - https://stackoverflow.com/questions/50890493/compile-error-when-stdfilesystem-header-file-is-added-to-my-program

I add more complex compiler checking:
```
// c++17: MinGW GCC version > 8.1
// c++17: Visual Studio 2017 version 15.7
#if ((__cplusplus >= 201703L) && ((!defined(__MINGW32__) && !defined(__MINGW64__)) || (__GNUC__ > 8))) || (defined(_MSVC_LANG) && (_MSVC_LANG >= 201703L))
```

SRombauts:
Hello, I believe that this is not needed anymore now that I merged the other ones, right?
I'll close it but let me know if something is still incorrect

Chronial:
Yes, this was already merged as part of #318

PR #334, opened 2021-07-25
We only need to link the cipher version of sqlite in case of HAS_CODEC

linux-fan-dave:
The current version of SQLiteCpp links the sqlcipher and the sqlite3 libraries when using codec. This leads to undefined behaviour since these two libraries provide the same symbols. In case of using codec only the sqlcipher library must be linked.


SRombauts:
Nice catch, thank you for taking the time to provide a fix!

PR #335, opened 2021-08-20
Disable std::filesystem on macOS when compiling C++17 but targetting <10.15

jagerman:
macOS flat out refuses to compile if you touch `std::filesystem::path` when targetting macOS < 10.15 (to be able to deploy a binary to older macOS versions) with the `-mmacosx-version-min=10.14` (or earlier) compiler flags.

If you try to compile code with `::path` (even if you don't actually call it) the Apple clang gods smite you with:
```
foo.cpp:8:26: error: 'path' is unavailable: introduced in macOS 10.15
        std::filesystem::path p;
                         ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.3.sdk/usr/include/c++/v1/filesystem:738:24: note: 'path' has been explicitly marked unavailable here
class _LIBCPP_TYPE_VIS path {
                       ^
```

This is reportedly caused by Apple's C++ team being somewhat incompetent at actually updating libc++ when they push out new releases, so libc++ stl features (in this case, to get at the std::filesystem exceptions from libc++) often lag behind the compiler by years, and Apple never releases libc++ updates for older mac versions because they don't care because C++ isn't Swift and only Swift is great and good.

\</rant\>

*Anyway*, this PR works around the issue by disabling the std::filesystem::path support when in C++17 mode when targetting macOS before 10.15.

I also broke up the hairy macro conditions because it was already hard to follow and would have gotten worse if kept in a single condition.

### DIFF
--- a/include/SQLiteCpp/Database.h
+++ b/include/SQLiteCpp/Database.h
@@ -14,9 +14,24 @@
 
 // c++17: MinGW GCC version > 8
 // c++17: Visual Studio 2017 version 15.7
-#if ((__cplusplus >= 201703L) && ((!defined(__MINGW32__) && !defined(__MINGW64__)) || (__GNUC__ > 8))) || (defined(_MSVC_LANG) && (_MSVC_LANG >= 201703L)) // NOLINT
+// c++17: macOS unless targetting compatibility with macOS < 10.15
+#if __cplusplus >= 201703L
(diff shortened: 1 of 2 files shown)


paulo-coutinho:
Please, can you check and merge?

SRombauts:
Yes, nice, thank you!

PR #339, opened 2021-10-04
Allow specifying transaction behaviors DEFERRED, IMMEDIATE, and EXCLUSIVE

jjenkins278:
See https://sqlite.org/lang_transaction.html.


SRombauts:
Thanks, that's a nice addition!

PR #344, opened 2021-11-22
Added std::vector version of getColumns()

Kacperos155:
I created a new variants of Statement::getColumns to retrive all columns objects of single row using one command:
- std::vector<Column> getColumns(const int aNumber = 0)
- std::vector<Column>& getColumns(std::vector<Column>& arBuffer, const int aNumber = 0)
- std::array<SQLite::Column, N> getColumns()
First two methods use std::vector to get whole row, when we don't know number of columns in compile time
or want to use a non const variable to retrieve columns.
Second method allows us to reuse std::vector, so we can allocate memory only once.
Third method is useful when we know number of columns in compile time and it is just an alias of
T getColumns(T, N).

This changes could simplify user codes, for examples:
- Retrieving information about columns in table
```cpp
SQLite::Statement q(...);
q.executeStep();
std::cout << "Columns: \n";
for ( const auto& column : q.getColumns() ) {
    std::cout << "Name: " << column.getName() << "\t Type: " << column.getType() << '\n';
}
```
- Processing whole table/query in range-based for loop
```cpp
SQLite::Statement q(...);
std::vector<SQLite::Column> row;
while( q.executeStep() ) {
    for ( const auto& column : q.getColumns(row) ) {
        std::cout << column.getText() << '\t';
    }
    std::cout << '\n';
}
```
- Deserializing big classes without using separate struct
```cpp
SQLite::Statement& operator<< (MyClass& c, SQLite::Statement& s)
{
    auto columns = s.getColumns<2>();
    c.m1 = columns[0].getUInt();
    c.m2 = columns[1].getText();
    return s;
}
```

I also made a test for this methods and renamed previous "getColumns" test to "getColumnsTemplate"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ example1
 *.a
 
 .vscode/
+.vs/
 /SQLiteCpp.sln
 *.ncb
 *.suo
--- a/include/SQLiteCpp/Statement.h
+++ b/include/SQLiteCpp/Statement.h
@@ -621,7 +621,7 @@ public:
      * @brief Return an instance of T constructed from copies of the first N columns
      *
      *  Can be used to access the data of the current row of result when applicable,
-     * while the executeStep() method returns true.
+     *  while the executeStep() method returns true.
      *
      *  Throw an exception if there is no row to return a Column from:
      * - if provided column count is out of bound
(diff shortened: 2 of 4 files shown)


Kacperos155:
This pull request was closed by accident when i was reorganizing my fork.
Now I will try to refactor whole Statement and Column classes.

I'm leaving changes of this pull request in branch [feature-getColumns](https://github.com/Kacperos155/SQLiteCpp/tree/feature-getColumns) in case when next changes would not get approved.

Issue #347, opened 2021-12-15
Feature request: move-assignment operator for Statement


Fytch:
(See #190)

The explicitly deleted copy-assignment operator of `SQLite::Statement` stops the compiler from generating a move-assignment operator for this type. On a related note, having a default constructor would also be nice. Consider the following example that I would wish to be able to write:

```cpp
struct foo
{
	SQLite::Database db;
	SQLite::Statement st;

	foo()
		: db(path, SQLite::OPEN_READWRITE)
	{
		// do stuff with db
		st = SQLite::Statement("SQL CODE");
	}
};
```
Currently, I have to box all statements that I use in this way into `optional`s. Ugly!

SRombauts:
Hello Fytch, thanks a lot for your good suggestion!

This should have been totally obvious to me when I switch to using C++11, but that wasn't long ago, and I have been busy with work & family.

Can you perhaps provide a Pull Request for it?
I should be able to get more time to work on this project this year, let's make it great again :)

PR #349, opened 2022-01-26
Refactoring of Statement and Column classes

Kacperos155:
As title says I refactored Statement and Columns with minimal changes to wrapper API.
List of all major changes:

- [x] **Replaced macro long size checks with _fixed width ints_ for Statement::bind**
(Macros are still needed for "operator long" in Columns but not in Statement binds methods. Fixed width ints should properly detect size of long function parameter)
- [x] **Replaced Statement::Ptr with std::shared_ptr<sqlite3_stmt>** (aliased as TStatementPtr) - need for this was mentioned [there](https://github.com/SRombauts/SQLiteCpp/blob/ce6dd9b82234e0427801380286d87f5bae417cca/include/SQLiteCpp/Statement.h#L712).
- [x] **Added Statement move assignment** (and so fixing #347 )
- [x] **Column throws when created with empty pointer to prepered statement.** 
- [x] **Statement methods: getIndex, getColumn, getExpandedSQL are now const**

### DIFF
--- a/include/SQLiteCpp/Column.h
+++ b/include/SQLiteCpp/Column.h
@@ -54,7 +54,7 @@ public:
      * @param[in] aStmtPtr  Shared pointer to the prepared SQLite Statement Object.
      * @param[in] aIndex    Index of the column in the row of result, starting at 0
      */
-    Column(Statement::TStatementPtr& aStmtPtr, int aIndex) noexcept;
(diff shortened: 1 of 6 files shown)


SRombauts:
This work was long overdue, I find it awesome that you tackled it and contributed back :)
Thanks!

PR #352, opened 2022-03-19
add basic meson support.

ninjaoflight:
This PR adds basic meson support, at the moment it does not compile on windows, however managed it to compile on Arch Linux, when the Windows support is ready it should fix also #53,#280 and add maybe put it on wrapdb fixing #97.

Issue #280, opened 2020-05-16
Running the tests on windows


SylvainCorlay:
Hi, many thanks for your work on this package. We use it for the [xeus-sqlite](https://github.com/jupyter-xeus/xeus-sqlite) project, a Jupyter kernel for SQLite, which you can try by clicking [here](https://mybinder.org/v2/gh/jupyter-xeus/xeus-sqlite/stable).

I am working on packaging SQLiteCpp on conda-forge for Windows (OS X and Linux builds are already availble).

- Our build at https://github.com/conda-forge/sqlitecpp-feedstock/pull/1 is succesful (https://github.com/conda-forge/sqlitecpp-feedstock/pull/1/commits/490b6c21a011178a890c1c3bba2c67068c710a71)
- But enabling the tests (https://github.com/conda-forge/sqlitecpp-feedstock/pull/1/commits/2ab08b0c5a34c078114e08dd98fce68e4eb25161) results in the following [failure](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=163139&view=logs&j=f97ca392-d626-52d0-4b2e-4b27aa319a77):

<details>

```
2020-05-16T09:50:21.4855913Z LINK: command "C:\PROGRA~2\MICROS~1\2017\ENTERP~1\VC\Tools\MSVC\1416~1.270\bin\Hostx64\x64\link.exe /nologo @CMakeFiles\SQLiteCpp_tests.dir\objects1.rsp /out:SQLiteCpp_tests.exe /implib:SQLiteCpp_tests.lib /pdb:%SRC_DIR%\build\SQLiteCpp_tests.pdb /version:0.0 /machine:x64 /INCREMENTAL:NO /subsystem:console SQLiteCpp.lib %PREFIX%\Library\lib\gtest-md.lib %PREFIX%\Library\lib\gtest_main-md.lib %PREFIX%\Library\lib\sqlite3.lib %PREFIX%\Library\lib\gtest-md.lib kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib /MANIFEST /MANIFESTFILE:SQLiteCpp_tests.exe.manifest" failed (exit code 1120) with the following output:
[...]

Issue #53, opened 2015-05-19
Build as a DLL ?


aurelien35:
Hi,

I plan to use SQLiteCpp in my project (windows).

Is there any reason to not build SQLiteCpp as a DLL ?

I didn't see any DLL import / export directives, and I prefer to use a DLL due to my project's architecture.

If you want, I can add the necessary macros.

SRombauts:
If you are willing to develop and test this for your need, I will integrate it, for sure!

ninjaoflight:
this was fixed in #406 feel free to test it

PR #354, opened 2022-04-20
Windows improved support (meson)

ninjaoflight:
this PR is an improvement of #352, managing to compile statically and run unit tests successfully on windows.
a summary of the changes in this PR:

- Meson file
  - use c++14 on windows(due to an error with xstddef)
  - add a preprocessor definition ('NON_AMBIGOUS_OVERLOAD') to use int instead of long to prevent ambiguous overload when compiling with clang++, it does not affect cmake and it only is enabled on meson
- Unit tests
  -  tests/Statement_test.cpp -> use "int64_t"(supported on c++11) instead of "long long" to prevent ambiguous overload errors (tested both on Arch Linux & Windows w clang)
  - tests/Statement_test.cpp -> check if NON_AMBIGOUS_OVERLOAD is defined and use int instead of long if is the case (to prevent ambiguous overload on windows)

the library still does not compile dynamically (as windows does not export the functions by default), so that should be fixed on an 3rd PR potentially fixing #280 and #53 for both meson and cmake, but this is a start before working on adding a macro for that feature.

after the dynamic compilation on windows is fixed, we could put the package on wrapdb so more users can use it easily.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -84,7 +84,14 @@ if not (host_machine.system() == 'windows' and cxx.get_id() == 'msvc')
         '-Wno-long-long',
     ]
 endif
-
+## using MSVC headers requires c++14, if not will show an error on xstddef as: 
+## 'auto' return without trailing return type; deduced return types are a C++14 extension
+if host_machine.system() == 'windows'
+    message('[WINDOWS] using c++14 standard')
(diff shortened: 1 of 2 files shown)


SRombauts:
Thanks a lot for this contribution.

Cheers!

PR #359, opened 2022-07-14
fix compilation issues earlier than iOS 13

panicfrog:
截屏2022-07-14 17 08 40


SRombauts:
Thanks for your contribution!

PR #367, opened 2022-09-13
add an option that could let the user decide whether to install SQLit…

setoye:
Hi:
First of all I want to say thanks for your library. It's pretty useful with elegant API design. I have encountered some issues when using this lib so I made this MR.
Currently I'm using a cmake based build system and trying to add `SQLiteCpp` into my project. Modern CMake has a lot of features to do that. like `add_subdirectory` or `FetchContent`. However both of them are not so good because of the install commands inside `SQLiteCpp`'s CMakeLists.txt. I don't want to install `SQLiteCpp` when invoking the install command after I build and install my projects. Some users, like me, may just want to static-link both `SQLiteCpp` and `sqlite3` internally. Besides, some users may just build `sqlite3` from source (`sqlite3` doesn't have an official cmake based build system as I know and that means they might write their own CMake scripts and create a `SQLite3` target, or even alias it as `SQLite::SQLite3`). So even if I set` SQLITECPP_INTERNAL_SQLITE` off, the install commands of `sqlite3` will cause a lot of CMake errors.
Hope this will be useful.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -310,30 +310,34 @@ target_include_directories(SQLiteCpp
     $<INSTALL_INTERFACE:include/>)
 
 # Allow the library to be installed via "make install" and found with "find_package"
+# An option that could let the user decide whether to install SQLiteCpp or not.
 
-include(GNUInstallDirs)
-install(TARGETS SQLiteCpp
-    EXPORT ${PROJECT_NAME}Targets
-    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
-    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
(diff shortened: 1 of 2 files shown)


SRombauts:
Hello, did you change your mind? Something is wrong with the proposal?

TimPushkin:
Wanted to use SQLiteCpp with the newer sqlite3 version (which I fetch myself with `FetchContent`) and also got errors related to the `install` commands of `SQLiteCpp`. Don't know why OP closed the PR, but the proposal is good.

TimPushkin:
An alternative is to create an `option` to make SQLiteCpp fetch sqlite by itself (I even found another closed PR with this: #240), but this would require bumping the CMake version to 3.14 to use `FetchContent`.

PR #377, opened 2022-11-21
Some documentation fixes

cbielow:
We have successfully used SQLiteCPP to transition from QtSql to your implementation with great success.
Thank you for developing SQLiteCPP :)

During our switch I came across some minor stuff, mostly concerning documentation, most of which is addressed in this PR.
In a few places I extended the docs to mention the underlying sqlite3 function, since it was not clear (to me) what was actually happening until I looked at the implementation. This may be different for experts in sqlite, but I only use it occasionally and found it more convenient if SQLiteCPP docs gave me as much information as possible.

Please let me know if this is desirable. I can revert things if you feel its too much detail.

One minor thing (which I did not fix): the documentation gets build in the source tree, instead of the build tree.
Is that intentional?

Thanks again for a great library!


SRombauts:
Thanks for contributing all these improvements :) 

About the documentation, I just set it up initially by copying from another of my C++ projects and didn't give it too many thoughts after that.
If it's useful to you and you have ideas on how it should be, feel free to propose :)

PR #379, opened 2022-12-06
Add support for C++ Unicode types

dougnazar:
Add support for using char8_t, char16_t, wchar_t, u8string, u16string & wstring when available.
wchar_t & wstring support are only available on platforms with a 2 byte wchar_t.

Usable to create Statements, bind parameters and retrieve column data.

The first commit is a fix for when the native database is in UTF-16 format and you try to extract a std::string.

### DIFF
--- a/include/SQLiteCpp/Column.h
+++ b/include/SQLiteCpp/Column.h
@@ -102,6 +102,49 @@ public:
      */
     std::string getString() const;
 
+#ifdef __cpp_unicode_characters
+    /**
+     * @brief Return a pointer to the text value (NULL terminated UTF-16 string) of the column.
+     *
+     * @warning The value pointed at is only valid while the statement is valid (ie. not finalized),
+     *          thus you must copy it before using it beyond its scope (to a std::u16string for instance).
+     */
+    const char16_t* getU16Text(const char16_t* apDefaultValue = u"") const noexcept;
+    /**
+     * @brief Return a std::u16string for a TEXT column.
+     *
+     * Note this correctly handles strings that contain null bytes.
+     */
+    std::u16string getU16String() const;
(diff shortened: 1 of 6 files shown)


SRombauts:
Hey hello @dougnazar,
I have merged only the first of your commits, the bugfix & unit tests, in order to release a 3.2.1 bugfix release ASAP without changes to APIs
Thanks

Pugnator:
@dougnazar what about char32_t and u32string? I'm using it al lot in asian language tools. Are you planning to create new PR with it or I'll do it instead when all your commits will be merged.
@SRombauts when are you planning to merge at least bind-related commits? I believe it's very useful feature.

dougnazar:
Unfortunately *SQLite* doesn't natively support UTF-32, so there is no API to pass through to. They've recently added `sqlite3_bind_text64()` however it only accepts `SQLITE_UTF8`, `SQLITE_UTF16`, `SQLITE_UTF16BE`, or `SQLITE_UTF16LE`.

I'm not sure that *SQLiteC++* is the correct layer to handle the conversion. It would add a *hidden* performance cost for using that API.

Pugnator:
> Unfortunately _SQLite_ doesn't natively support UTF-32, so there is no API to pass through to. They've recently added `sqlite3_bind_text64()` however it only accepts `SQLITE_UTF8`, `SQLITE_UTF16`, `SQLITE_UTF16BE`, or `SQLITE_UTF16LE`.
> 
> I'm not sure that _SQLiteC++_ is the correct layer to handle the conversion. It would add a _hidden_ performance cost for using that API.

I agree, will stick to external 32-16 conversion until it's natively supported. Thank you for your enhancements!

PR #380, opened 2022-12-08
[Meson] fixes for meson project

UnixY2K:
## PR Description

This PR does the following things.
- .gitignore: ignore `.cache` directory created by vscode when using clangd
- Meson specific files
    -  increment the project version to `3.2.0`
    -  increase gtest version to 1.12.1 and use the version that includes main(only applies when gtest is not installed/detected so it can compile out of the box in meson)
    - add missing comentaries for the meson options

> Note for Windows: there is a current bug in meson/ninja when using  clang see mesonbuild/meson#10022, the workaround is setting up manually the `CXX` and `CXX_LD` env vars manually to be `clang++` and `lld-link` if clang is installed

if there is anything else required please let me know

## Future aditions on another PR

- Add a meson setup section in the readme so any new user can set it up easily if not familiar with meson
- Add dynamic link support in meson/windows targets (already done on a private fork but requires some changes)
- Put a public a public wrap file in meson wrapdb so any user can install using `meson wrap install sqlitecpp`, this is out of scope for this repo, but should be done once the dynamic windows support is ready.

If there is anything else missing please let me know

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -93,6 +93,11 @@ if host_machine.system() == 'windows'
     sqlitecpp_opts += [
         'cpp_std=c++14',
     ]
+    # check that we are not trying to build as dynamic library
+    if get_option('default_library') != 'shared'
+        warming('SQLiteCpp does not support shared library on Windows, the library will be built as static')
+    endif
+    
 endif
 # Options relative to SQLite and SQLiteC++ functions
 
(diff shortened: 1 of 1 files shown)


dougnazar:
I have a local patch here I've been meaning to send to make a meson install match a cmake install. I guess most meson users are just bundling/wrapping rather than installing. Feel free to include in your future PR if you want since you're working on that area.

```
diff --git a/meson.build b/meson.build
index fb6f658..e1bdc2e 100644
--- a/meson.build
+++ b/meson.build
@@ -146,7 +146,7 @@ libsqlitecpp = library(
     dependencies: sqlitecpp_deps,
     # override the default options
     override_options: sqlitecpp_opts,
-    # install: true,
+    install: true,
     # API version for SQLiteCpp shared library.
     version: '0',)
 if get_option('SQLITECPP_BUILD_TESTS')
@@ -173,6 +173,9 @@ install_headers(
     'include/SQLiteCpp/Transaction.h',
     'include/SQLiteCpp/VariadicBind.h',
     'include/SQLiteCpp/ExecuteMany.h',
+    'include/SQLiteCpp/Savepoint.h',
+    'include/SQLiteCpp/Utils.h',
+    subdir : 'SQLiteCpp',
 )

 sqlitecpp_dep = declare_dependency(
```

ninjaoflight:
> I have a local patch here I've been meaning to send to make a meson install match a cmake install. I guess most meson users are just bundling/wrapping rather than installing. Feel free to include in your future PR if you want since you're working on that area.
> 
> ```
> diff --git a/meson.build b/meson.build
> index fb6f658..e1bdc2e 100644
> --- a/meson.build
> +++ b/meson.build
> @@ -146,7 +146,7 @@ libsqlitecpp = library(
>      dependencies: sqlitecpp_deps,
>      # override the default options
>      override_options: sqlitecpp_opts,
> -    # install: true,
> +    install: true,
>      # API version for SQLiteCpp shared library.
>      version: '0',)
>  if get_option('SQLITECPP_BUILD_TESTS')
> @@ -173,6 +173,9 @@ install_headers(
>      'include/SQLiteCpp/Transaction.h',
>      'include/SQLiteCpp/VariadicBind.h',
>      'include/SQLiteCpp/ExecuteMany.h',
> +    'include/SQLiteCpp/Savepoint.h',
> +    'include/SQLiteCpp/Utils.h',
> +    subdir : 'SQLiteCpp',
>  )
> 
>  sqlitecpp_dep = declare_dependency(
> ```

thanks for the notice!
I added the missing savepoint for the install_headers, it was also missing from headers and sources (and added the subdir option which was missing), I would need to test it later on a non windows machine
as for the install option I plan to add it after I finish the updates to the windows required changes(that will be on other PR), once I have that done I will uncomment the option on that PR, but if I forgot feel free to remember me so I can modify it.

eli-schwartz:
You can install the whole directory with Meson too:

```meson
install_subdir('include/SQliteCpp', install_dir: get_option('includedir'))
```

ninjaoflight:
I think we can use the second option of installing the subdir, so we don´t need to keep track of the files but let users choose if they want to install them

thanks for the replies!

meson.build
```diff
@@ -161,20 +161,9 @@ if get_option('SQLITECPP_BUILD_TESTS')
...
-    'include/SQLiteCpp/Exception.h',
-    'include/SQLiteCpp/Savepoint.h',
-    'include/SQLiteCpp/Statement.h',
-    'include/SQLiteCpp/Transaction.h',
-    'include/SQLiteCpp/VariadicBind.h',
-    'include/SQLiteCpp/ExecuteMany.h',
-    subdir : 'SQLiteCpp',
-)
+install_subdir(
+    'include/SQliteCpp',
```
dougnazar, line 165: s/SQliteCpp/SQLiteCpp/

ninjaoflight: done

SRombauts:
Thanks a lot for the outstanding teamwork, reviews, and patches :)

SRombauts:
I think it's also time to release a new version, probably just a patch 3.2.1, that will come with all the little cleanup & build system update that already accumulated.
What do you think?

ninjaoflight:
sure, since Meson support on Windows is already good, we can also set it as the first version in Meson Wrapdb.
I don´t think that the dynamic link support on windows should be an issue since the static compilation workaround is good enough for most workloads

Issue #378, opened 2022-11-24
Make use of std::filesystem optional via e.g. cmake


timosachsenberg:
I think it could make sense to make the use of filesystem functions optional.
In our case we don't use them, but it causes a few build errors on different systems.
E.g., we can't support older macOS clang versions and also the gcc on Ubuntu 18.04 failed to
properly detect whether we have the experimental/filesystem or filesystem header.

BTW, for us, we got it working by adding these two small lines as a patch to Database.h
https://github.com/OpenMS/OpenMS/pull/6498/files

but maybe the use of filesystem is something that could also be configured through cmake?

piperoc:
Thanks for the fix, got me out of a jam:  I need to build for Ubuntu 18.04 and 20..04 so your fix helped a ton.
I agree that a CMake option might be cleaner.

ninjaoflight:
I think it could be possible to add a option that adds a preprocesor definition that disables it, it should be something like this:
```cmake
option(SQLITECPP_DISABLE_STD_FILESYSTEM "Disable the use of std::filesystem in SQLiteCpp." OFF)
if (SQLITECPP_DISABLE_STD_FILESYSTEM)
    message (STATUS "Disabling std::filesystem support")
    add_definitions(-DSQLITECPP_DISABLE_STD_FILESYSTEM)
endif (SQLITECPP_DISABLE_STD_FILESYSTEM)
```
and then under database.h after `SQLITECPP_HAVE_STD_FILESYSTEM` and `SQLITECPP_HAVE_STD_EXPERIMENTAL_FILESYSTEM`
```cpp
#ifdef SQLITECPP_DISABLE_STD_FILESYSTEM
    #undef SQLITECPP_HAVE_STD_FILESYSTEM
    #undef SQLITECPP_HAVE_STD_EXPERIMENTAL_FILESYSTEM
#endif
```
so we can manually disable the support and detection of std::filesystem

another option is to use the `__has_include` macro of c++17 like the following after the other checks for std::filesystem

```cpp
#ifdef __has_include
    #if __has_include(<filesystem>)
        #define SQLITECPP_HAVE_STD_FILESYSTEM
    #else
        #undef SQLITECPP_HAVE_STD_FILESYSTEM
    #endif
    #if __has_include(<experimental/filesystem>)
        #define SQLITECPP_HAVE_EXPERIMENTAL_FILESYSTEM
    #else
[...]

PR #388, opened 2022-12-11
make std::filesystem optional

ninjaoflight:
## Description
Add a switch to disable `std::filesystem` so it can be disabled if required on older compilers, this should let users that older compilers versions to disable it, solving #378 for meson and cmake, additionally checks in the source code if the header is found

## changes
- CMakeLists.txt: adds a new option that is disabled by default
- Database.h
  -  adds an additional check for `<filesystem>` and `<experimental/filesystem>`
  - checks for the previous switch and disables the support if enabled
- `meson.build`/`meson.options` adds the same option that disables `std::filesystem` support

I think this PR should be made on `3.2.1` or greater as is a new feature/enhancement itself and should come with the other planned project cleanups, please let me know if there is anything else required

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -294,6 +294,13 @@ else (SQLITECPP_INTERNAL_SQLITE)
     endif()
 endif (SQLITECPP_INTERNAL_SQLITE)
 
+## disable the optional support for std::filesystem (C++17)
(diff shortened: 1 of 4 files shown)


ninjaoflight:
closed by mistake

SRombauts:
Thanks!
And yes I agree, this should be released as part of 3.2.1, which I plan to release tomorrow if all goes well.

Sadly, there were no CI/CD checks on this PR, I think because it was closed (or I missed something else?) so I'll have to check status on master post-merge.

Cheers!

PR #390, opened 2022-12-12
fix incorrect work of savepoint

spoyler:
Add release for savepoint after rollabck in destructor for deleting it form sqlite transaction stack.


SRombauts:
Thanks for providing this :) 

I think the main issue is that the old unit test didn't show that there was a bug, and now it doesn't prove either that the bug is fixed.
Could you add a test for that? It's not urgent but it would definitely help ensure it works as expected, and also show how it is expected to behave

SRombauts:
OK, I merged it without the unit tests.

I should rename the rollback() method to rollbackTo() now that I understand better the differences!

Thanks again

PR #391, opened 2022-12-13
meson project changes based on wrap submission review

ninjaoflight:
## description

this PR makes checks if there is a bundled sqlite3 in OSX that does not support load extension and disables the support automatically

> note: line feeds were changed from CRLF to LF

## changes
- meson_options.txt
  - changed `SQLITE_OMMIT_LOAD_EXTENSION` to `SQLITE_OMIT_LOAD_EXTENSION` (typo fix)
- meson. Build
  - change the version to 3.2.1
  - changed `SQLITECPP_USE_STACK_PROTECTOR` to `SQLITECPP_USE_STACK_PROTECTION` (typo fix)
  - adds the `SQLITE_OMIT_LOAD_EXTENSION` option and enables it if running on MacOS with a bundled sqlite that does not supports it

## additional notes

on wrapdb this can be added as a patch as is only a change in the meson files and not the project itself

> edit: there is a list of other changes that can be seen on the wrapdb repo

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 project(
     'SQLiteCpp', 'cpp',
     # SQLiteCpp requires C++11 support
-    default_options: ['cpp_std=c++11'],
+    default_options: ['cpp_std=c++11', 'warning_level=3'],
     license: 'MIT',
     version: '3.2.1',
 )
@@ -51,8 +51,9 @@ sqlitecpp_srcs = files(
 sqlitecpp_args = cxx.get_supported_arguments(
     # included in meson by default
     # -Wall
-    '-Wextra',
-    '-Wpedantic',
+    # included when warning_level=3
+    #'-Wextra',
+    #'-Wpedantic',
     '-Wswitch-enum',
     '-Wshadow',
     '-Wno-long-long',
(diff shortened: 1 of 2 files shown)


ninjaoflight:
note: related to mesonbuild/wrapdb#816

ninjaoflight:
currently doing a code review for this update on wrapdb repo

SRombauts:
Hello @ninjaoflight

Thank you again for all the work you are doing in this area!

Could you write a quick start on how to use meson? I think it's worth adding to a section in the readme, after the chapter on CMake.

Then I should add a new Github Actions workflow to test meson build automatically

meson.build
```diff
@@ -39,18 +39,24 @@ sqlite3_dep = dependency(
...
-]
-sqlitecpp_args = [
-    '-Wall',
-]
+)
+sqlitecpp_args = cxx.get_supported_arguments(
+    # included in meson by default
+    # -Wall
+    '-Wextra',
+    '-Wpedantic',
```
eli-schwartz, line 55: Wextra and Wpedantic are in meson too, simply set the `default_options: ['warning_level=3']`

eli-schwartz (changes requested):

meson.build
ninjaoflight: changed

PR #392, opened 2022-12-13
Support different encryption libraries

jowr:
By default, SQLiteCpp support SQLCipher for encrypting a database and it is very difficult to integrate other SQLite3 libraries. The changes from this PR improve this situation and make it easier to integrate other SQLite3 implementations in the future.

The code has been tested with [SQLite3MultipleCiphers](https://github.com/utelle/SQLite3MultipleCiphers) which has to be built separately. It can be found afterwards by setting the `SQLITE3MULTIPLECIPHERS_DIR` variable in SQLiteCpp or within the project that includes SQLiteCpp.

Please do not hesitate to ask for further changes.


jowr:
> * check that `sqlcipher` is included and define `SQLITECPP_USE_SQLITE3MULTIPLECIPHERS` if enabled
> * check that `SQLite3MultipleCiphers` is included and define `SQLITECPP_USE_SQLCIPHER` if enabled

Actually, it is the other way round - `SQLITECPP_USE_SQLITE3MULTIPLECIPHERS` relates to using `SQLite3MultipleCiphers` and `SQLITECPP_USE_SQLCIPHER` switches to `sqlcipher`.

Furthermore the user should set the options `SQLITECPP_USE_SQLITE3MULTIPLECIPHERS` or `SQLITECPP_USE_SQLCIPHER` to express their intention of using one or the other. The compile flags are then set accordingly. Both libraries also set the `SQLITE_HAS_CODEC` compile flag. By default, the two encryption libraries are disabled and the internal SQLite version is enabled via the option `SQLITECPP_INTERNAL_SQLITE`. 

> * check that cannot be more than one encryption library at the same time

That is correct. I use the alias target `SQLite::SQLite3` to determine whether SQLite3 has been found. The script throws an error if the target exists and you try to enable another SQLite3 implementation.

In case all implementations are disabled, the script tries to find the default `SQLite::SQLite3` on your system.

ninjaoflight:
> > * check that `sqlcipher` is included and define `SQLITECPP_USE_SQLITE3MULTIPLECIPHERS` if enabled
> > * check that `SQLite3MultipleCiphers` is included and define `SQLITECPP_USE_SQLCIPHER` if enabled
> 
> Actually, it is the other way round - `SQLITECPP_USE_SQLITE3MULTIPLECIPHERS` relates to using `SQLite3MultipleCiphers` and `SQLITECPP_USE_SQLCIPHER` switches to `sqlcipher`.
> 

thanks, yes that was a typo from my side.

> By default, the two encryption libraries are disabled and the internal SQLite version is enabled via the option SQLITECPP_INTERNAL_SQLITE

as for the libraries, is it required to use the internal implementation or there is no issue if we use the system provided?, mainly because on meson the dependencies are handled differently so the implementation might be somewhat different if that is the case

jowr:
No, there should be no problem using the system SQLite3 as long as you disable the internal implementation. This behaviour is unchanged compared to how it works today.

image

Pichas:
Hi, @jowr. I am using SQLite3MultipleCiphers as well. So I want to notice that you have a memory leak with this impl of SQLiteCPP. You also need to apply this patch to fix it. SQLite3MultipleCiphers alocates additional stuff when you use enription

```diff 
# ./src/Database.cpp

diff --git a/src/Database.cpp b/src/Database.cpp
index 320695f..9d4f38d 100644
--- a/src/Database.cpp
+++ b/src/Database.cpp
@@ -14,7 +14,7 @@
 #include <SQLiteCpp/Backup.h>
 #include <SQLiteCpp/Exception.h>
 #include <SQLiteCpp/Statement.h>
-
+#include <sqlite3mc.h>
 #include <sqlite3.h>
 #include <fstream>
 #include <string.h>
@@ -90,6 +90,7 @@ void Database::Deleter::operator()(sqlite3* apSQLite)
     // Only case of error is SQLITE_BUSY: "database is locked" (some statements are not finalized)
     // Never throw an exception in a destructor :
     SQLITECPP_ASSERT(SQLITE_OK == ret, "database is locked");  // See SQLITECPP_ENABLE_ASSERT_HANDLER
+    sqlite3_shutdown();
 }
 
 // Set a busy handler that sleeps for a specified amount of time when a table is locked.
```

SRombauts:
I am closing this PR in favor of the related one #532 that I have finally just merged

PR #393, opened 2022-12-13
Fix preprocessor issues

jowr:
- Do not compare the version number for external SQLite3 libraries
- Correct use of the __has_include macro to stop the preprocessor from complaining


SRombauts:
Thanks, I noticed the new warning and made the same change locally!

PR #394, opened 2022-12-14
check if SQLITE_OPEN_NOFOLLOW is defined

ninjaoflight:
## description

under MacOS 11 BigSur(`macos-latest`) the macro `SQLITE_OPEN_NOFOLLOW` is not defined so it may cause compilation issues for those versions, you can check more about it on this repo which extracts it: https://github.com/ninjaoflight/macos-latest-checks

this seems to be an edge case as the sqlite version on that image is reported as `3036000` but does not adds it, I don´t know if this issue is related to only the MacOS GitHub Image, however this should also fix those edge corners where this modification is present.

this was noticed under the meson review, so this could be added on a separate parch under wrapdb, which has not been merged as this issue is present, however this should not been an issue for any future CI run as `mac-os latest` will be MacOS 12 Monterey

please let me know if there is anything else required


SRombauts:
Hey, thanks for this. I agree that it's strange, and I didn't want to rush it yesterday, but it cannot do any harm after all :)

PR #399, opened 2022-12-20
add disable option for sqlite3_expanded_sql

### LINKED ISSUE #213: SQLiteCpp/build/libSQLiteCpp.so: undefined reference to `sqlite3_expanded_sql'
I am receiving only one sqlite3 undefined reference error after adding the -lsqlite3 flag. But this single error has been preventing g++ from building.

ninjaoflight:
## Description
Add an option to disable `sqlite3_expanded_sql` that disables the internal use of it inside the method `Statement::getExpandedSQL` so it gives a warning when building SQLiteCpp, at runtime it will give an exception instead

this should fix #213

## changes
- CMakeLists.txt: add the `SQLITECPP_DISABLE_EXPANDED_SQL` option along with the define `SQLITECPP_DISABLE_EXPANDED_SQL`
- meson.build/meson.options: add the same  `SQLITECPP_DISABLE_EXPANDED_SQL`  that defines `SQLITECPP_DISABLE_EXPANDED_SQL`
- Statement.cpp
  - give a warning at compile time if  `SQLITECPP_DISABLE_EXPANDED_SQL` is not defined when running the version of sqlite < 3.14.0
  - give an exception when trying to run `Statement::getExpandedSQL` if compiled witouth support for `sqlite3_expanded_sql`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -304,6 +304,13 @@ if (SQLITECPP_DISABLE_STD_FILESYSTEM)
     add_definitions(-DSQLITECPP_DISABLE_STD_FILESYSTEM)
 endif (SQLITECPP_DISABLE_STD_FILESYSTEM)
 
+## disable the optional support for sqlite3_expanded_sql (from sqlite3 3.14.0)
+option(SQLITECPP_DISABLE_EXPANDED_SQL "Disable the use of sqlite3_expanded_sql in SQLiteCpp." OFF)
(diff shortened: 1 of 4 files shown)


ninjaoflight:
I think that we could also add an special exception like `SQLite::UnsupportedMethodException` for any method that does not want to expose the internal implementation of sqlite3 and not make use of the same Exception used for internal SQLite3 Exceptions

SRombauts:
Thanks

PR #400, opened 2022-12-20
Lint corrections

ninjaoflight:
## Description
make simple corrections in both build systems (Meson/CMake)
- remove the use of `add_definitions` for `std::filesystem` support in CMake
- fix the definition of `SQLITE_HAS_CODEC` in Meson


SRombauts:
PR/issue 400, that's a milestone!

PR #401, opened 2022-12-21
Fix meson installation

dougnazar:
Fix include directory name so include files get installed. 
Flag library for installation on non-Windows platforms.


SRombauts:
Thanks

Issue #264, opened 2020-01-23
Add documentation for prepared statements in transactions


bigOconstant:
The readme shows an example of prepared statements and transactions.(Both very helpful) No examples are given showing prepared statements within transactions. If it is possible to have prepared statements within transactions could someone add an example in the readme of such a scenario?

PR #404, opened 2023-01-04
Add documentation for prepared statements in transactions

ewarchul:
I've encountered the same problem with prepared statements and transactions as described in https://github.com/SRombauts/SQLiteCpp/issues/264, and I thought it would be good to resolve this (relatively) minor issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -373,6 +373,42 @@ catch (std::exception& e)
 }
 ```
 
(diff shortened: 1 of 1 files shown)


SRombauts:
Thanks,
I think that we have much more examples in code and unit tests, but this way it's more visible.

Issue #403, opened 2023-01-03
[Meson] CI for meson project


ninjaoflight:
## Description

Happy new Year!

As Meson support is becoming more stable it would be good to have a CI for the meson project so we can check any bug or edge case like #394 or #378 this issue should be a good starting point in order to define how the CI should look like along with any feedback that is required in order to complete it, additionally having this CI will help us add any new feature that may not be included con CMake or Meson itself so it can be fixed and tested carefully before merging

an initial proposed format should look like the example of the first comment
as you may notice it is pretty similar to the CI that is already in CMake, but needs to be checked/reviewed first in order to check if everything is working good before opening a PR.

please let me know what do you think or if you have any idea/sugestion.

ninjaoflight:
```yaml
name: build

on: [push, pull_request]

jobs:
  build:
    name: ${{ matrix.config.name }}
    runs-on: ${{ matrix.config.os }}
    strategy:
      fail-fast: false
      matrix:
        config:
        - {
            name: "Windows Latest MSVC",
            os: windows-latest,
            generator: "Visual Studio 17 2022",
            build_type: "Debug",
            cc: "cl", cxx: "cl",
            extra_path: "",
          }
        - {
            name: "Windows Latest MinGW",
            os: windows-latest,
            generator: "MinGW Makefiles",
            build_type: "Debug",
            cc: "gcc", cxx: "g++",
            extra_path: "C:\\ProgramData\\chocolatey\\lib\\mingw\\tools\\install\\mingw64\\bin",
          }
        - {
            name: "Ubuntu Latest GCC",
            os: ubuntu-latest,
            generator: "Unix Makefiles",
            build_type: "Debug",
            cc: "gcc", cxx: "g++",
            extra_path: ""
          }
        - {
            name: "macOS Latest Clang",
            os: macos-latest,
            generator: "Unix Makefiles",
            build_type: "Debug",
            cc: "clang", cxx: "clang++",
            extra_path: ""
          }

    steps:
    - uses: actions/checkout@v2
    - name: submodule
      run: git submodule update --init --recursive
    - name: extra_path
      shell: bash
      run: echo "${{matrix.config.extra_path}}" >> $GITHUB_PATH
    - name: configure
      shell: cmake -P {0}
      run: |
        set(ENV{CC} ${{matrix.config.cc}})
        set(ENV{CXX} ${{matrix.config.cxx}})
    - name: install prerequisites
      run: |
        # asuming that python and pip are already installed
[...]

PR #384, opened 2022-12-11
Add Mingw GitHub actions

SRombauts:
Fix #382 #74 disable SQLITECPP_USE_STACK_PROTECTION when running on MinGW

PR #407, opened 2023-02-10
Create Meson CI

ninjaoflight:
PR that completes #403, adding the base meson CI pipeline

### DIFF
--- a/.github/workflows/meson.yaml
+++ b/.github/workflows/meson.yaml
@@ -48,7 +48,7 @@ jobs:
           }
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: submodule
       run: git submodule update --init --recursive
     - name: extra_path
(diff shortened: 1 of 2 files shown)


ninjaoflight:
getting (`0xC0000139`) on both `Windows MSVC` and `Windows Clang`, which appears to be the same `DLL not found`. I think it's the same `MinGW` error from #384, need to check this later and check which DLLs are missing

ninjaoflight:
> getting (`0xC0000139`) on both `Windows MSVC` and `Windows Clang`, which appears to be the same `DLL not found`. I think it's the same `MinGW` error from #384, need to check this later and check which DLLs are missing

seems that the sqlite3.dll was not found, managed to fix it by building it instead via `--force-fallback-for=sqlite3`

ninjaoflight:
fixed the macOS CI error by using the fallback SQLite3

ninjaoflight:
Fixed Clang errors by fixing the setup of environment variables, as for MSVC used a task that enabled the automatic discovery of MSVC

SRombauts:
Thanks a lot for all this! God job

PR #408, opened 2023-02-12
correct executable name in meson

ninjaoflight:
change the name to be `SQLITECPP_sample_demo2` instead of `SQLITECPP_sample_demo1`


SRombauts:
Thanks

PR #410, opened 2023-03-05
use checkout@v3 in CMake CI

ninjaoflight:
There is a warning due to the usage of checkout@V2 that can be fixed by using V3 instead.


SRombauts:
Thanks

PR #413, opened 2023-04-01
Fix compiler warnings warning

### LINKED ISSUE #412: Getting `‘visibility’ attribute ignored` warnings
When SQLiteCpp is built with GCC, a lot of warnings like these arise:
```
/home/runner/work/SQLiteCpp/SQLiteCpp/src/Column.cpp:21:25: warning: ‘visibility’ attribute ignored [-Wattributes]
   21 | SQLITECPP_API const int INTEGER   = SQLITE_INTEGER;
      |                         ^~~~~~~
/home/runner/work/SQLiteCpp/SQLiteCpp/src/Column.cpp:22:25: warning: ‘visibility’ attribute ignored [-Wattributes]
   22 | SQLITECPP_API const int FLOAT     = SQLITE_FLOAT;
      |                         ^~~~~
/home/runner/work/SQLiteCpp/SQLiteCpp/src/Column.cpp:23:25: warning: ‘visibility’ attribute ignored [-Wattributes]
   23 | SQLITECPP_API const int TEXT      = SQLITE_TEXT;
      |                         ^~~~
/home/runner/work/SQLiteCpp/SQLiteCpp/src/Column.cpp:24:25: warning: ‘visibility’ attribute ignored [-Wattributes]
   24 | SQLITECPP_API const int BLOB      = SQLITE_BLOB;
      |                         ^~~~
/home/runner/work/SQLiteCpp/SQLiteCpp/src/Column.cpp:25:25: warning: ‘visibility’ attribute ignored [-Wattributes]
   25 | SQLITECPP_API const int Null      = SQLITE_NULL;
      |                         ^~~~
```

The above is copied from SQLiteCpp's own CI on Ubuntu.

ninjaoflight:
# Fix several compiler warnings

fixed several warnings that come from #406, most of the warnings are not harmful but it is better to fix them
- removed the  `#pragma message` inside `SQLiteCppExport.h`  as it showed as a warning
- removed `SQLITECPP_API` inside cpp files as it will be ignored and show a warning instead
- [meson only] check if stack protection is supported before adding the flag for it

this should fix #412

### DIFF
--- a/include/SQLiteCpp/SQLiteCppExport.h
+++ b/include/SQLiteCpp/SQLiteCppExport.h
@@ -21,10 +21,8 @@
 #if defined(_WIN32)&& !defined(__GNUC__) && defined(SQLITECPP_COMPILE_DLL)
     #if SQLITECPP_DLL_EXPORT
         #define SQLITECPP_API __declspec(dllexport)
-        #pragma message("Exporting symbols")
     #else
         #define SQLITECPP_API __declspec(dllimport)
-        #pragma message("Importing symbols")
     #endif    
 #else    
     #if __GNUC__ >= 4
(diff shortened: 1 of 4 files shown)


SRombauts:
Hey @ninjaoflight, for some reason I was certain I had already merged this one.
I am sorry, and thank you again for all your work!

PR #415, opened 2023-04-11
Remove mismatched else condition in CMakeLists.txt

Timmmm:
From the CMake docs:

> Per legacy, the else() and endif() commands admit an optional <condition> argument. If used, it must be a verbatim repeat of the argument of the opening if command.

Doesn't really matter but it could be confusing to people who aren't very familiar with CMake's intricacies. Also it triggers an annoying lint warning in my IDE.


SRombauts:
Thanks!

PR #424, opened 2023-05-29
Set Meson version to 3.3.0

ninjaoflight:
Sets the meson package version to 3.3.0


SRombauts:
Thanks, I somehow missed this one

PR #431, opened 2023-06-12
Add 6 Statement::bind() overrides using size_t for blobs

### LINKED ISSUE #430: Binding BLOB requires narrowing conversion
```
const int lengthInBytes =
      static_cast<int>(sizeof(uint8_t) * binary->raw().size());
insert.bind(1, binary->raw().data(), lengthInBytes);
```

Would be nice if the API took size_t or something so that I don't have to narrow it.
On my machine size_type (size_t) is unsigned long.

SRombauts:
Fix #430 for a better support of standard containers

Also add a couple of tests (not covering all variants)

### DIFF
--- a/include/SQLiteCpp/Statement.h
+++ b/include/SQLiteCpp/Statement.h
@@ -156,6 +156,10 @@ public:
      * @note Uses the SQLITE_TRANSIENT flag, making a copy of the data, for SQLite internal use
      */
     void bind(const int aIndex, const void*         apValue, const int aSize);
+    void bind(const int aIndex, const void*         apValue, const size_t aSize)
+    {
+        bind(aIndex, apValue, static_cast<int>(aSize));
+    }
     /**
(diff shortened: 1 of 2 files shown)


hellow554:
I am currently using this great library. Thank you for this!

One issue I'm running into is binding a blob breaks the variadic binding possiblity, because it requires two parameters, not one.

So I would like to introduce a new struct `Blob` or similar, that wraps a void pointer and `std::size_t`, so I can use it in a variadic binding suitation, e.g.:

```cxx
// [...]
struct Blob {
    void *ptr;
    std::size_t size;
};
SQLite::bind(st, "ABC", 2, SQLite::Blob{"def", 3});
```

I think this would perfectly fit in this PR, what do you think?

SRombauts:
I am going to release a new version, and then I'll update the baseline c++ versions, make the code cleaner and allow breaking APIs

PR #448, opened 2023-08-28
[Meson] update meson dependencies

ninjaoflight:
this PR only affects meson, however a later change could be made for CMake
updates the following depenencies:
 - GTests to 1.14.0
 - sqlite to 3.42.0
makes 
sets the recommended version to C++14 as Gtest now requires C++14, additionally shows a warning and sets it to c++14 if
- C++11 is set and is on windows (this was #354 but now checks if is C++11 and shows a warning instead of setting it to C++14 so it does not revert when using a newer version)
- C++11 is set and uses [GTest 1.14.0](https://github.com/google/googletest/releases/tag/v1.14.0) or greater,  for more information check [Google's Foundational C++ Support Policy](https://opensource.google/documentation/policies/cplusplus-support)

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,8 @@
 project(
     'SQLiteCpp', 'cpp',
-    # SQLiteCpp requires C++11 support
-    default_options: ['cpp_std=c++11', 'warning_level=3'],
+    # SQLiteCpp supports C++11
+    # however newer versions of gtest requires C++14
+    default_options: ['cpp_std=c++14', 'warning_level=3'],
     license: 'MIT',
     version: '3.3.1',
 )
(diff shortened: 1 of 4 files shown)


ninjaoflight:
getting an error on TravisCI
for the error in Windows latest clang seems to be due to a compiler version

ninjaoflight:
for windows-latest (clang) is caused due to a older version of clang, we could ignore it as it will be fixed in actions/runner-images#8125

SRombauts:
Hello,
Yes, I noticed errors with Clang. It was possible to retry and get rid of it in the last few days, but not anymore (probably updating their images progressively)
And don't worry about Travis CI; it was blocked for months, saying I overused it and I was going to get rid of it... But I finally unlocked my account, and I now need to fix the outdated workflow

SRombauts:
Hello @ninjaoflight, I fixed Travis CI with https://github.com/SRombauts/SQLiteCpp/pull/450 🎉

SRombauts:
And it's merged, thank you!

PR #469, opened 2024-06-18
Explicitly =delete; Statement::bindNoCopy(..., std::string&&)

maysl:
Rvalues are inherently unsuitable for no-copy binding, because their lifetime cannot be guaranteed. Separately declare, and delete, all overloads of bindNoCopy() that take a std::string rvalue.

### DIFF
--- a/include/SQLiteCpp/Statement.h
+++ b/include/SQLiteCpp/Statement.h
@@ -179,6 +179,11 @@ public:
      */
     void bindNoCopy(const int aIndex, const void*           apValue, const int aSize);
     /**
+     * @brief Deleted, because the value's lifetime could not be guaranteed. Use bind().
+     */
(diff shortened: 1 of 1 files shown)


maysl:
Hi @SRombauts,

We ran into the issue that in a long sequence of legitimate `bindNoCopy()` calls, there was one that used a `std::to_string()` as its value. This sort of mistake can easily happen, but it's equally easy to mitigate. Hence the PR.

Thank You for your work!

UnixY2K:
seems that the CI issue in macos is caused due to [PEP 668](https://peps.python.org/pep-0668/), I will make a PR to fix the issue

UnixY2K:
created #475 to fix the macOS CI issue

SRombauts:
Thanks a lot for your contribution. 
Sébastien

PR #475, opened 2024-08-14
Update meson dependencies [Meson only]

UnixY2K:
makes the following changes
- update meson wrap dependencies
  - update gtest 1.14.0 -> 1.15.0
  - update sqlite 3.42.0 -> 3.46.1
- removed redundant variables that pointed to the same sources in example executables

### DIFF
--- a/examples/example1/meson.build
+++ b/examples/example1/meson.build
@@ -1,8 +1,5 @@
 
-
-example1_sources = files(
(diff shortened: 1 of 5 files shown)


UnixY2K:
there is an issue with macOS CI that should be fixed in #476

SRombauts:
Thanks!

PR #476, opened 2024-08-14
Macos ci fix

UnixY2K:
fixes a CI issue in OSX due to [PEP-668](https://peps.python.org/pep-0668/)
pip3 was used to install meson and ninja build, pipx is a drop in replacement for this use case.


SRombauts:
Thanks a lot for your contribution!
Sébastien

PR #488, opened 2024-09-07
SQLiteCpp/Statement.h: add missing `<cstdint>` include

trofi:
Without the change the build fails on `gcc-15` as:

    [  5%] Building CXX object CMakeFiles/SQLiteCpp.dir/src/Backup.cpp.o
    In file included from /build/source/include/SQLiteCpp/Column.h:14,
                     from /build/source/include/SQLiteCpp/Database.h:14,
                     from /build/source/include/SQLiteCpp/Backup.h:15,
                     from /build/source/src/Backup.cpp:12:
    /build/source/include/SQLiteCpp/Statement.h:132:39: error: 'uint32_t' does not name a type
      132 |     void bind(const int aIndex, const uint32_t      aValue);
          |                                       ^~~~~~~~
    /build/source/include/SQLiteCpp/Statement.h:20:1: note: 'uint32_t' is defined in header '<cstdint>'; this is probably fixable by adding '#include <cstdint>'
       19 | #include <memory>
      +++ |+#include <cstdint>
       20 |

### DIFF
--- a/include/SQLiteCpp/Statement.h
+++ b/include/SQLiteCpp/Statement.h
@@ -14,6 +14,7 @@
 #include <SQLiteCpp/Exception.h>
 #include <SQLiteCpp/Utils.h> // SQLITECPP_PURE_FUNC
 
+#include <cstdint>
 #include <string>
 #include <map>
 #include <memory>


SRombauts:
Thanks a lot for taking the time to contribute these two fixes, much appreciated

PR #489, opened 2024-09-07
tests/Database_test.cpp: fix a warning around `#endif`

trofi:
Without the change `gcc` detects unexpected tokens as:

    /build/source/tests/Database_test.cpp:566:9: warning: extra tokens at end of #endif directive [8;;https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wendif-labels-Wendif-labels8;;]
      566 | #endif  && !defined(__APPLE__)
          |         ^~


SRombauts:
Thanks!

PR #494, opened 2024-10-08
[docs] update meson instructions

UnixY2K:
makes the following changes to the readme:
- fold some sections into summary.
- mention the usage of wrapdb.
- remove the mention of the local wrap file that no longer exists.


UnixY2K:
the failing CI is not related to this PR so it can be safely ignored, only changes in the readme were made

Issue #495, opened 2024-12-09
CMake: SQLITE_OMIT_LOAD_EXTENSION not properly set


brt-v:
I'm fairly new to this project, so apologies if I'm overlooking something.

The main CMakeList.txt offers a `SQLITE_OMIT_LOAD_EXTENSION` options and sets it for the `SQLiteCpp` project ([here](https://github.com/SRombauts/SQLiteCpp/blob/master/CMakeLists.txt#L252)). However, it does not set it for `[sqlite3](https://github.com/SRombauts/SQLiteCpp/tree/master/sqlite3)`

So I think that in the `sqlite3` [CMakeLists.txt](https://github.com/SRombauts/SQLiteCpp/blob/master/sqlite3/CMakeLists.txt) this should be added:
```
if (SQLITE_OMIT_LOAD_EXTENSION)
    target_compile_definitions(sqlite3 PUBLIC SQLITE_OMIT_LOAD_EXTENSION)
    message(STATUS "Compile sqlite3 with SQLITE_OMIT_LOAD_EXTENSION")
endif (SQLITE_OMIT_LOAD_EXTENSION)
```

Also, a related question: the sqlite documentation lists some [Recommended Compile-time Options](https://www.sqlite.org/compile.html). Can (some of) these be set in conjunction with SQLiteCpp? What's the best way to do this with CMake?

Thanks!

UnixY2K:
I think for the first you can submit a PR to get it merged, however you need to check if the found sqlite is the current bundled.

for the second option the same applies if using the bundled sqlite + having an option to change the value (default would be true)

brt-v:
Thanks, I just submitted a PR.

> however you need to check if the found sqlite is the current bundled.

the new option is only applied to the bundled `sqlite3`, so I don't think there's an issue for the case that it isn't bundled.

> for the second option the same applies if using the bundled sqlite + having an option to change the value (default would be true)

This makes sense if an sqlite compile option affects SQliteCpp, but if it's doesn't it's quite cumbersome to add an option for every available SQlite setting. 

But luckily there's a simpler solution: I can just call `target_compile_definitions(sqlite3 PUBLIC ...)` directly from my own CMakeLists.txt. I just gave it a quick try with example1 and the following CMakeLists.txt. This works for me (Windows/VS2022):
```cmake
cmake_minimum_required(VERSION 3.12)

project(SQLiteCpp_ex1)

[...]

PR #496, opened 2024-12-14
sqlite3: set SQLITE_OMIT_LOAD_EXTENSION

brt-v:
Addresses #495.


SRombauts (approved):
Great, thanks for taking the time to propose that!

PR #504, opened 2025-05-19
Update googletest to v1.17.0

SRombauts:


SRombauts:
problem with gtest: https://github.com/SRombauts/SQLiteCpp/pull/504

PR #509, opened 2025-05-23
fix: faulty undef

SGSSGene:
In some cases a non existing define is being undefined. This fixes this case.
SQLITECPP_HAVE_EXPERIMENTAL_FILESYSTEM => SQLITECPP_HAVE_STD_EXPERIMENTAL_FILESYSTEM


SRombauts:
Thanks!

Issue #510, opened 2025-05-23
Uptodate Platforms


SGSSGene:
Hey,
I have seen that the target platforms are for example Ubuntu 14.04, 16.04 and 18.04 (Travis CI and Github Actions).
I believe this is an outdated list. Is it possible to update the list?

PR #511, opened 2025-05-23
fix: update the c++ version selection process

SGSSGene:
The c++ version should be set via `target_compile_features(SQLiteCpp PUBLIC cxx_std_11)`
This only sets this as a minimum version for this target, and not all targets.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@
 #
 # Distributed under the MIT License (MIT) (See accompanying file LICENSE.txt
 # or copy at http://opensource.org/licenses/MIT)
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.8)
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake") # custom CMake modules like FindSQLiteCpp
 project(SQLiteCpp VERSION 3.3.3)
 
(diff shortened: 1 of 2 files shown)


SRombauts:
Hey @SGSSGene, that sounds like a nice improvement as well, thanks for taking the time to provide the PR!

CMakeLists.txt
```diff
@@ -213,6 +204,9 @@ endif()
 # add sources of the wrapper as a "SQLiteCpp" static library
 add_library(SQLiteCpp ${SQLITECPP_SRC} ${SQLITECPP_INC} ${SQLITECPP_DOC} ${SQLITECPP_SCRIPT})
 
+# set minimum version to c++11
+target_compile_features(SQLiteCpp PUBLIC cxx_std_11)
```
SRombauts, line 208: The cxx_std_11 compiler meta-feature is not available CMake 3.5 (which was the minimal version until now, see line 7 cmake_minimum_required(VERSION 3.5))

It was introduced in 3.8, [CMake 3.8 release notes](https://cmake.org/cmake/help/v3.8/release/3.8.html#c-c)

Could you please also update the minimum version?
Thanks!

SGSSGene:
@SRombauts: All changes that we currently do downstream in the OpenMS project :-)

What CMake version is suitable?
This is also why I asked here, what the current supported platforms are https://github.com/SRombauts/SQLiteCpp/issues/510
I would like to bump the CMake version up, as far possible.

SRombauts:
> I would like to bump the CMake version up, as far possible.
My strategy up to know regarding supported platforms has always be to try to keep it backward compatible as long as possible.
Her for instance I would just update from 3.5 to 3.8.

Then, if you have any other pull request with more changes requiring another bump, I would evaluate that again.
But if you have any specific version in mind with a good reason I am happy to consider it instead.

SRombauts:
Hello, I am going to consider this ASAP after the next release: a good cleanup and major version upgrade!

PR #512, opened 2025-05-23
fix: hide installation behind an option

SGSSGene:
Currently, every project pulling in SQLiteCpp automatically gets an `make install` which installs SQLiteCpp into the system. This is often not desired, since other projects only statically link. The solution is to add another "SQLITECPP_INSTALL" option, which allows to turn on/off this behavior.


SRombauts:
Thanks, that's indeed a nice improvements :)

PR #513, opened 2025-05-29
Add support for C++ modules

mikomikotaishi:
This pull request adds support for C++20 modules to SQLiteCpp, by creating modules `sqlite.*`. It is built using the macro `SQLITECPP_BUILD_MODULES`. The `README.md` file has also been updated to contain information.

I also added a `namespace sqlite = SQLite;` alias so that the namespace and module names align.


mikomikotaishi:
I am wondering, could we also add a preprocessor macro to export the symbol `Exception` as `SQLException` (similar to its name in Java), so that it is not confused with the usual `std::exception` from the standard library?

SRombauts:
This would be something for C++20, right now I am only considering making C++17 the baseline...
I'll have a second though at that in the following days

mikomikotaishi:
It is an opt-in feature, and doesn't prevent C++17; it just makes the option available if the user compiles as C++20 and above, because the configuration is guarded in CMake.

SRombauts:
Noted!
I am going to release a final version supporting C++11 to help overhaul the entire project, so this is something to keep in mind for right after

Issue #517, opened 2025-08-03
std::string_view support


dvtate:
I frequently use [`std::string_view`](https://en.cppreference.com/w/cpp/string/basic_string_view.html) to avoid copying sub-strings. I wanted to pass a string_view to `Statement::bindNoCopy` but saw the type error. My first instinct was to replace it with `statement.bindNoCopy(1, sv.data(), sv.size())` but after checking the implementation, I noticed that it calls `sqlite3_bind_blob` when I would have expected it to call `sqlite3_bind_text`.

## Proposal
I'm sure one of the following approaches would work.

### Support binding std::string_view
Example implementation. I would also like to use Statment::bindNoCopy with string_view.
```cpp
#if __cplusplus >= 201703L // C++17
// Bind a string value to a parameter "?", "?NNN", ":VVV", "@VVV" or "$VVV" in the SQL prepared statement
void Statement::bind(const int aIndex, const std::string_view& aValue)
{
    const int ret = sqlite3_bind_text(getPreparedStatement(), aIndex, aValue.data(),
                                      static_cast<int>(aValue.size()), SQLITE_TRANSIENT);
    check(ret);
}
#endif
```

<details>
	<summary>Research</summary>

 According to [the sqlite3 C interface documentation](https://www.sqlite.org/c3ref/bind_blob.html) it seems that `sqlite3_bind_text` doesn't require that the given `const char*` parameter be null-terminated since we also pass a size argument that's not -1. So we should be free to use std::string_view here without problems.
	
> In those routines that have a fourth argument, its value is the number of bytes in the parameter. To be clear: the value is the number of bytes in the value, not the number of characters. If the fourth parameter to sqlite3_bind_text() or sqlite3_bind_text16() is negative, then the length of the string is the number of bytes up to the first zero terminator. If the fourth parameter to sqlite3_bind_blob() is negative, then the behavior is undefined. If a non-negative fourth parameter is provided to sqlite3_bind_text() or sqlite3_bind_text16() or sqlite3_bind_text64() then that parameter must be the byte offset where the NUL terminator would occur assuming the string were NUL terminated. If any NUL characters occur at byte offsets less than the value of the fourth parameter then the resulting string value will contain embedded NULs. The result of expressions involving strings with embedded NULs is undefined.

</details>

Due to implicit casting, this function could actually replace the following two overloads when string_view is available.
- `void bind(const int aIndex, const std::string&  aValue);`
[...]

PR #355, opened 2022-05-24
Supports string_view for bind

xis19:
With this patch, it is possible to bind `string_view`, which addresses PR #298.

PR #224, opened 2019-09-26
Added some bindings for std::string_view when using c++17

kaptenhonek:
Title says it all :p

PR #518, opened 2025-08-04
std::string_view support

dvtate:
My goal is to fix this issue: https://github.com/SRombauts/SQLiteCpp/issues/517

Previous, similar PRs:
 - https://github.com/SRombauts/SQLiteCpp/pull/224 : ambiguous overloads
 - https://github.com/SRombauts/SQLiteCpp/pull/355 : missing C++17 ifdef guards
 
 I should probably add unit tests and examples using string_view.

### DIFF
--- a/include/SQLiteCpp/Column.h
+++ b/include/SQLiteCpp/Column.h
@@ -16,6 +16,9 @@
 
 #include <string>
 #include <memory>
+#if __cplusplus >= 201703L // C++17
+#include <string_view>
+#endif
 
 // Forward declarations to avoid inclusion of <sqlite3.h> in a header
 struct sqlite3_stmt;
@@ -102,6 +105,18 @@ public:
      * Note this correctly handles strings that contain null bytes.
      */
     std::string getString() const;
+
+#if __cplusplus >= 201703L // C++17
+    /**
+     * @brief Return a std::string_view for a TEXT or BLOB column.
+     *
+     * Note this correctly handles strings that contain null bytes.
+     *
+     * @warning returned string_view is only valid until there is a type
+     *          conversion or the statement is stepped or reset.
+     */
+    std::string_view getStringView() const;
+#endif
 
     /**
      * @brief Return the type of the value of the column using sqlite3_column_type()
--- a/include/SQLiteCpp/Database.h
+++ b/include/SQLiteCpp/Database.h
@@ -18,7 +18,6 @@
 // c++17: macOS unless targetting compatibility with macOS < 10.15
 #ifndef SQLITECPP_HAVE_STD_EXPERIMENTAL_FILESYSTEM
 #if __cplusplus >= 201703L
-    #include <string_view>
     #if defined(__MINGW32__) || defined(__MINGW64__)
         #if __GNUC__ > 8 // MinGW requires GCC version > 8 for std::filesystem
             #define SQLITECPP_HAVE_STD_FILESYSTEM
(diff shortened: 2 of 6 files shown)


jagerman:
@dvtate - it would be nice to also see this solve #533 where string_view is not only nice to have but actually critically important for an application that wants to use secure memory for encrypted database keys.

jagerman:
> does this look like something you think would make sense for the project?

I know you didn't ask me, but personally I find this PR highly valuable: currently this library has *no ability* to bind TEXT values into statement placeholders without copying, which feels like a big omission versus the C API, which allows this easily.

I tried to work around the limitation by seeing if I could get the raw C API statement handle out of a SqliteCpp::Statement so that I could bind such values myself via the C API, but that proved impossible: Statement's internal handle is private and thus entirely inaccessible and so my choices were either: 1) avoid SQLite::Statement entirely and use the C API for all binding if I want to be able to do copy-less TEXT binding; 2) give up and just accept unnecessary copying; 3) fork the project to add it.

I opted for 3, but luckily found this PR already written and so am currently building on a fork with this merged.

(The key changes here are much less important, because Database *does* expose the C handle, so there it's pretty simple to just go to the C API for that one call).

dvtate:
@SRombauts bump

SRombauts:
Hey @dvtate, thanks a lot for the proposal and the bump!
I agree that we should really have this feature added, it's really a good idea.

I am sorry for the long delay, but sadly the PR was opened at a time when I was in vacation, the subject was too complex to wrap my head around, and then life happened :sweat: 

Now, on the implementation, I see a few important issues that we should address before merging:

1. First (and I know it's my own fault for the delay), a rebase on master would be really needed.
2. There are too many changes: it would be much better to focus on just the original intent, and I am going fix #533 in a separate PR -> https://github.com/SRombauts/SQLiteCpp/pull/571 
3. We need to cover the code with unit tests.
4. We should discuss some of the changes related to ABI compat (it's broken now, not sure if it's really a big deal) and the C++11 / C++ 17 compat.

Let me know if you are willing to handle that, I would understand if it's too much work, esp. given how unresponsive I have been. 

Sébastien

dvtate:
> Let me know if you are willing to handle that

Unless something comes up, I should have enough free time this week.

Knocking out 1-3 should be straightforward.

For 4, Because `std::string_view` is a C++17 feature I put those functions in `#if` blocks. This is done already for `std::filesystem` support, so I assumed this is okay. Only thing I can see getting in the way is implicit conversions.

Given that this is a minor, performance-oriented feature, vacation should definitely take priority IMO. Glad to see you're back tho :)

dvtate:
Ok, I re-based my changes on master.

I'm seeing now that the library is built with C++11 so the string view methods don't get compiled. To avoid changing that I'll have to add in some C++11-compatible helper methods to isolate std::string_view to the headers.

SRombauts:
And there is the compatibility issue I was talking about: mixing c++11 & c++17. We want to be able to build the lib in one version, and use it in multiple clients with different version

Something that we should do for (or before) this task is adding a few of CI config to test : 

1. build SQLiteCpp in c++11 but a "consummer" app (the tests or the example) in c++17
2. (the opposite is probably not an option)
3. both in c++17
4. both in c++11

In summary, it would mean perhaps just building 2 "consumers" apps with different c++ levels, to prevent any regression.

SRombauts:
Unrelated, but, `Column::getStringView()` could probably be header-only using the existing `getBlob()` / `getBytes()` C++11 API.

But be careful with empty values: `sqlite3_column_blob()` may return `nullptr`, so you would need to return an empty `std::string_view{}` explicitly rather than constructing directly from the returned pointer, leading to undefined behaviro (I think) with `string_view(nullptr, 0)`

dvtate:
> `Column::getStringView()` could probably be header-only

Added that in commit.

I added some private `Statement::bindText64` and `Statement::bindTextNoCopy64` helpers to implement the string_view code to the headers. It seems that there's a lot of code that relies on implicit `char*` -> `void*` conversion for blobs, otherwise I would have made them public overloads.

With this change, the library can be built with C++11 and used with C++17 users. It appears the tests are built with C++17 as I'm seeing my string_view tests (within macros) passing.

It appears that `std::string_view(nullptr, 0)` is specified as UB... I'll fix that.

SRombauts:
Thanks a lot for your hard work, I see that the CI is failing... I am going to review all that asap, either tonight or tomorrow.

I am actually considering removing support for C++11: I should release one last version and upgrade SQLiteCpp to a new version 4.0.0 with a more recent CMake requirement to support C++17 as the baseline!

This should make these changes simpler, and could help us modernize and simplify the codebase

dvtate:
Oops, looking at the CI logs I forgot to put the column tests behind c++17 macros. I'll fix that.

jagerman:
> I should release one last version and upgrade SQLiteCpp to a new version 4.0.0 with a more recent CMake requirement to support C++17 as the baseline!

:+1: on this idea: C++11 is pretty ancient at this point and C++17 is more or less available everywhere and brings a lot of nice language and STL improvements.

jagerman:
Or to put it another way:

> Version 3.0.0 - January 31 2020
>     C++11 is now required

That was 8 years 11 months after C++11 finalized.  C++17 finalized 9 years 6 months ago.  (And this makes me feel old).

jagerman:
> It appears that `std::string_view(nullptr, 0)` is specified as UB... I'll fix that.

That's not UB: that's exactly what a default-constructed string_view does.  (It's construction from a nullptr without a size, or with a non-zero size, that are UB).

dvtate:
Yea, as long as C++11 projects can still use the library it should be fine. C++11 is like Java 8 -- many are afraid to bump the number even if they don't have to use any of the new features.

Assuming the library and consumer are using the same/similar compilers everything should be fine afaik.

dvtate:
@jagerman 
> That's not UB

Agreed that there's no reason for it to be UB but cppreference had some spooky wording and I'm not a language lawyer.

image
image

 From what I'm seeing online,` [(char*)nullptr, (char*)nullptr+0)` is a valid empty range... so probably fine. 

[This SO thread](https://stackoverflow.com/a/79677748) suggests that it's not UB but a lot of the sources they reference relate to `std::basic_string` and not specifically `std::basic_string_view`.

So it's probably safe to remove the check. Unless there's a remote possibility that `getBlob()` returns `nullptr` and `getBytes() != 0`.

SRombauts:
Regarding the potential UB I was referring to, the confusion was coming from the following:

```
std::string_view(nullptr, 0); // OK in C++17
std::string_view(nullptr);    // UB in C++17: tries traits::length(nullptr)
```

So you can just forget about it :pray:

SRombauts:
I have been trying to do a thorough review of your code changes, that looks good overall, but I have had a few questions to dig deeper (including on some of my own codes that I couldn't remember why it was changed later, namely in `Column::getString()` where I have adjusted the comment on master)
I'll have to take more time later to write down a few minor comments

dvtate:
Because of the code there I assumed I couldn't do
```c++
int len = getBytes();
return std::string_view((const char*) getBlob(), len);
```

Take your time reviewing.

PR #536, opened 2026-03-14
use std namespace for int types under cstdint

UnixY2K:
integer types under `cstdint` are only required to be under the `std` namespace while the global namespace definition is optional.

this PR fixes the undefined behavior for the files that assume global namespace availability.


GuillaumeDua (approved):
LGTM 👍 Thanks

